PR #1, opened 2016-10-31
Bug 1313154 - adjust beetmoverscript to cope with nightly graph constraints

MihaiTabara:
@lundjordan r?

Milestone - proper bits into place to tweak beetmoverwork to bind nicely with taskgraph as we discussed last week. I thought it'd be useful to land this first to unblock @kmoir while I work on a separate branch to massage the manifest for balrogworker.  

We should now be able to schedule beetmover tasks for signed and unsigned bits, at least for Fennec.

I have few leftovers left to address later:
* add more tests to ensure 100% coverage
* docstrings + docs everywhere
* add Travis support
* make Travis deploy automatically to Pypi every time we tag a new release


beetmoverscript/script.py
```diff
@@ -30,16 +31,30 @@
     context.task = get_task(context.config)  # e.g. $cfg['work_dir']/task.json
     # 2. validate the task
     validate_task_schema(context)
-    # 3. generate manifest
+    # 3 grab manifest props with all the useful data
+    context.properties = await get_props(context)
```
lundjordan: does this need to be async? I think since this only downloads one file and everything up till now is sequential, I'm not sure if we will see a benefit from concurrency here.

On the other hand, I suppose scriptworker could be doing background things and would appreciate the non-blocking io download. Plus if you removed the async/await, you would need to use something other than retry_request. So I don't think it's worth addressing.

beetmoverscript/script.py
```diff
@@ -30,16 +31,30 @@
...
+    # 4. generate manifest
     manifest = generate_candidates_manifest(context)
-    # 4. for each artifact in manifest
+    # 5. for each artifact in manifest
     #   a. download artifact
     #   b. upload to candidates/dated location
     await move_beets(context, manifest)
-    # 5. copy to releases/latest location
+    # 6. copy to releases/latest location
```
lundjordan: I think this logic has bitrotted and is part of move_beets. Feel free to remove the line.

beetmoverscript/templates/fennec_nightly_en_us_multi_signed.yml
```diff
@@ -1,12 +1,4 @@
----
-metadata:
-    name: "Beet Mover Manifest"
-    description: "Maps Fennec Nightly artifacts to pretty names for the en-US and multi locale"
-    owner: "release@mozilla.com"
-
-artifact_base_url: https://queue.taskcluster.net/v1/task/{{ taskid_to_beetmove }}/artifacts/public/build/
-s3_prefix_dated: pub/mobile/nightly/{{ upload_date }}-mozilla-central-android-api-15/
-s3_prefix_latest: pub/mobile/nightly/latest-mozilla-central-android-api-15/
+{% include "fennec_nightly_en_us_multi_base.yml" %}
```
lundjordan: nice call 👍

beetmoverscript/utils.py
```diff
@@ -15,21 +15,44 @@ def load_json(path):
         return json.load(fh)
 
 
+def infer_template_args(context):
+    props = context.properties
+    tmpl_key_option = "signed" if context.task["payload"]["update_manifest"] is True else "unsigned"
```
lundjordan: nice. this could be confusing to outsiders. maybe put a comment how we determine whether this is a signed task or not?

beetmoverscript/utils.py
```diff
@@ -15,21 +15,44 @@ def load_json(path):
...
     payload = context.task['payload']
+
     template_args = {
-        "version": payload['version'],
-        # payload['upload_date'] is a timestamp defined by params['pushdate'] in mach taskgraph
-        "upload_date": arrow.get(payload['upload_date']).format('YYYY/MM/YYYY-MM-DD-HH-mm-ss'),
-        "taskid_to_beetmove": payload["taskid_to_beetmove"]
+        "taskid_to_beetmove": payload["taskid_to_beetmove"],
+        "taskid_of_manifest": payload["taskid_of_manifest"],
+        "update_manifest": payload["update_manifest"],
```
lundjordan: do we need update_manifest passed to the yml template file?

beetmoverscript/utils.py
```diff
@@ -15,21 +15,44 @@ def load_json(path):
...
 
+def infer_template_args(context):
+    props = context.properties
+    tmpl_key_option = "signed" if context.task["payload"]["update_manifest"] is True else "unsigned"
+
+    return {
+        "version": props["appVersion"],
+        "branch": props["branch"],
+        "product": props["appName"],
+        "stage_platform": props["stage_platform"],
```
lundjordan: do we need stage_platform or product passed to the yml template file?

beetmoverscript/script.py
MihaiTabara: Heh :) Would you believe me if I told you that initially that was my workaround? That was my motivation as well and it was all working fine but then I ran the PR on a short feedback session with async-@rail who asynced everything and passed me the virus to async everything 😃 

On a more serious note, your second statement is the motivation that stands behind async-ing that call there. I added requests as dep and used requests.get there before spotting the retry_request in scriptworker and async-ing the stuff.

beetmoverscript/utils.py
MihaiTabara: Yep, you're right. I'll address this now here but I'll add a separte issue on this repo to document everything. I left docstrings, travis, tests, coverage and others for when we have a stable working version.

PR #7, opened 2016-11-07
Bug 1313154 - adjust beetmoverworker to cope with dekstop nightlies. …

MihaiTabara:
…r=jlund

### DIFF
--- a/beetmoverscript/script.py
+++ b/beetmoverscript/script.py
@@ -21,7 +21,7 @@ from scriptworker.utils import (retry_async, download_file,
 from beetmoverscript.constants import MIME_MAP, MANIFEST_URL_TMPL, PLATFORM_MAP
 from beetmoverscript.task import validate_task_schema
 from beetmoverscript.utils import (load_json, generate_candidates_manifest,
-                                   amend_props)
+                                   update_props)
 
 log = logging.getLogger(__name__)
 
@@ -34,13 +34,13 @@ async def async_main(context):
     validate_task_schema(context)
     # 3 prepare manifest props file
     #   a. grab manifest props with all the useful data
-    #   b. amend its platform field to proper one
+    #   b. amend platform field to proper one
     context.properties = await get_props(context)
-    amend_props(context, PLATFORM_MAP)
+    context.properties = update_props(context.properties, PLATFORM_MAP)
     # 4. generate manifest
     manifest = generate_candidates_manifest(context)
     # 5. for each artifact in manifest
-    #   a. download artifac
+    #   a. download artifact
     #   b. upload to candidates/dated location
     await move_beets(context, manifest)
 
--- a/beetmoverscript/script.py
+++ b/beetmoverscript/script.py
@@ -18,9 +18,10 @@ from scriptworker.exceptions import ScriptWorkerTaskException, ScriptWorkerRetry
 from scriptworker.utils import (retry_async, download_file,
                                 raise_future_exceptions, retry_request)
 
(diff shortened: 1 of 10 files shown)


MihaiTabara:
@lundjordan r?

* couldn't find a better way for now to remap those platforms to be consistent with production
* once we have the signed CMARs, I'll remove the currently CMAR from unsigned tasks

beetmoverscript/utils.py
```diff
@@ -66,3 +66,13 @@ def generate_candidates_manifest(context):
...
     return manifest
+
+
+def amend_props(context, platform_mapping):
+    """Function to alter the `stage_platform` field from balrog_props to their
+    corresponding correct values for certain platforms"""
+
+    _platform = context.properties["stage_platform"]
+    # for some products/platforms this mapping is not needed, hence the default
+    context.properties["stage_platform"] = platform_mapping.get(_platform,
```
lundjordan: maybe we shouldn't overwrite `stage_platform` and instead save the new platform prop as just `platform`? That way it's clear that we have two variants of the platform name? I see in `infer_template_args` you already rename it to just `platform`

beetmoverscript/templates/firefox_nightly_en_us_signed.yml
```diff
@@ -0,0 +1,20 @@
...
+{% for locale in ['en-US'] %}
+  {{ locale }}:
+  {% if platform == "linux-i686" %}
+    package:
+      artifact: target.tar.bz2
+      s3_key: firefox-{{ version }}.{{ locale }}.{{ platform }}.tar.bz2
+
+  {% endif %}
+
+  {% if platform == "linux-x86_64" %}
```
lundjordan: are linux-x86_64 and linux-i686 broken into two blocks because we plan on adding more artifacts that might differ? Or should we say `if platform in ['linux-x86_64', 'linux-i686']` ?

beetmoverscript/script.py
```diff
@@ -31,12 +32,15 @@
     context.task = get_task(context.config)  # e.g. $cfg['work_dir']/task.json
     # 2. validate the task
     validate_task_schema(context)
-    # 3 grab manifest props with all the useful data
+    # 3 prepare manifest props file
+    #   a. grab manifest props with all the useful data
+    #   b. amend its platform field to proper one
     context.properties = await get_props(context)
+    amend_props(context, PLATFORM_MAP)
```
lundjordan: nit: I'm trying to get myself in the habit of writing pure code with no side-effects. could you make amend_props a non mutating method?

e.g. `context.properties = update_props(context.properties)` where in update_props body impl you deepcopy context.properties and return the updated copy

MihaiTabara: Really liked this comment and suggestion 👍 I wasn't pretty happy about my workaround either but somehow I moved forward. But "no-side effects" is a good desideratum so I'll gladly comply ;)

beetmoverscript/script.py
```diff
@@ -31,12 +32,15 @@
...
     # 4. generate manifest
     manifest = generate_candidates_manifest(context)
     # 5. for each artifact in manifest
-    #   a. download artifact
+    #   a. download artifac
```
MihaiTabara: busted :)

beetmoverscript/utils.py
MihaiTabara: Cool, will do so!

MihaiTabara:
@lundjordan Super useful comments, thanks again!
Always useful to have a second pair of eyes glancing at the code ;)

PR #9, opened 2016-11-08
Bug 1313154 - generate beetmover output to be used by balrogworker, r?mihai

lundjordan:
@MihaiTabara r?

I took your revs from https://github.com/MihaiTabara/beetmoverscript/pull/1 , resolved conflicts, and added some new behaviour:

1. addressed my own comments from original pr
2. adds manifest to artifacts
3. adds ability to specify which beets are to be submitted instead of every beet from signed task
4. adjust manifest props based on changes made here: https://github.com/mozilla-releng/funsize-balrogworker/pull/2

I tested this locally (commented out uploading for speed reasons) and it seems to behave as expected:

task: https://tools.taskcluster.net/task-inspector/#AA7X9PepRQWzZWa3FZHpkw/
manifest.json https://public-artifacts.taskcluster.net/AA7X9PepRQWzZWa3FZHpkw/0/public/manifest.json


beetmoverscript/script.py
```diff
@@ -19,15 +19,18 @@
...
 from beetmoverscript.utils import (load_json, generate_candidates_manifest,
-                                   update_props)
+                                   update_props, get_hash)
 
 log = logging.getLogger(__name__)
 
 
 # async_main {{{1
 async def async_main(context):
+    # balrog_manifest is used by a subsequent balrogworker task that points to a beetmoved artifact
```
MihaiTabara: I guess it makes more sense to always declare this. Much cleaner, good call.

beetmoverscript/script.py
```diff
@@ -43,14 +46,18 @@
     #   a. download artifact
     #   b. upload to candidates/dated location
     await move_beets(context, manifest)
-
+    # 6. write balrog_manifest to a file and add it to list of artifacts
+    if context.task["payload"]["update_manifest"]:
```
MihaiTabara: In the attempt to ditch `update_manifest` variable, here we could substitue with something like:

if context.balrog_manifest:
         add_balrog_manifest_to_artifacts(context)

since `balrog_manifest` will remain an empty list if no task from manifest has the  `update_balrog_manifest` flag set. However, we'd still need to use the `update_manifest` var when we toggle between the unsigned/signed templates.

beetmoverscript/script.py
```diff
@@ -69,15 +76,17 @@
                                       manifest['mapping'][locale][deliverable]['s3_key'])
             dest_latest = os.path.join(manifest["s3_prefix_latest"],
                                        manifest['mapping'][locale][deliverable]['s3_key'])
+            balrog_manifest = manifest['mapping'][locale][deliverable].get('update_balrog_manifest')
```
MihaiTabara: chapeau bas for this cool idea. way more neater to flag them from the manifest, rather than add more logic and if-checks for which artifacts to add in the `manifest.json`. Smart move, good job!

beetmoverscript/script.py
```diff
@@ -86,6 +95,46 @@
     await retry_download(context=context, url=source, path=abs_file_path)
     await retry_upload(context=context, destinations=destinations, path=abs_file_path)
 
+    if update_balrog_manifest:
+        context.balrog_manifest.append(
```
MihaiTabara: Really appreciated the pure non-mutable approach. Good follow-up from yesterday's conversation 👍

beetmoverscript/script.py
```diff
@@ -86,6 +95,46 @@
...
+            enrich_balrog_manifest(context, abs_file_path, locale, destinations)
+        )
+
+
+def enrich_balrog_manifest(context, path, locale, destinations):
+    props = context.properties
+
+    if props["branch"] == 'date':
```
MihaiTabara: I'd suggest we move all the hardcoded values like `date` in constants file to ease any potential tweaking later on if we ever change the dev branch. Or even add another one. But I think it's totally fine for now.

beetmoverscript/script.py
```diff
@@ -86,6 +95,46 @@
...
+                                         s3_key=destinations[0])
+        url_replacements = []
+    else:
+        # we extract the dated destination as the 'latest' is useless
+        url = "{prefix}/{s3_key}".format(prefix="https://archive.mozilla.org",
+                                         s3_key=destinations[0])
+        url_replacements = [['http://archive.mozilla.org/pub, http://download.cdn.mozilla.net/pub']]
+
+    return {
+        "tc_fennec_nightly": True,
```
MihaiTabara: We'll need to address this once we're done with Fennec and switch are attention for desktop Firefox nightlies.

beetmoverscript/script.py
lundjordan: hm, er, what am I talking about.

in the payload we could revert back to have: context.task["payload"]["template_key"] which just states the explicit template key.

then, yes, at this line we go with what you suggest `if context.balrog_manifest:`

beetmoverscript/script.py
lundjordan: yeah, I didn't like hardcoding anything branch specific. thanks for pointing this out.

beetmoverscript/script.py
lundjordan: lol. I blame no brainz

beetmoverscript/script.py
MihaiTabara: Added this as an issue to keep this in mind.

beetmoverscript/script.py
MihaiTabara: Added it as a reminder under https://github.com/mozilla-releng/beetmoverscript/issues/11

beetmoverscript/script.py
MihaiTabara: Added as a reminder in https://github.com/mozilla-releng/beetmoverscript/issues/12

PR #15, opened 2016-11-22
Bug 1318037 - adjust beetmover to cope with l10n. r=Callek

MihaiTabara:
r? @Callek 

Will follow-up with puppet patch to fix the script_config.json in beetmoverworker too and to wrap all this up in a python package.


Callek:
Looks good, nits:

.DS_Store should go away (add it to gitignore and we'll be fine, windows at least creates it automagically)
checksums on desktop repacks won't actually be valid, with or without CoT, but not a big deal imo for now.

MihaiTabara:
@Callek I removed the .DS_STore, added it by accident.
I'll let those checksums for now - I have TODOs and open-issues to tackle them anyway before going Tier-1.

MihaiTabara:
@Callek Added it in the .gitignore too.
Please approve the PR once you're happy with it.

PR #26, opened 2017-01-08
adds cot support by relying on upstreamArtifacts for beets

lundjordan:
for an example task.json and payload with upstreamArtifacts, see https://tools.taskcluster.net/task-inspector/#YkE_OQjoTj2G92P4A4Cy1A/ which was ran against a local beetmoverscript

two notes about YkE_OQjoTj2G92P4A4Cy1A: 
  1. since my bandwidth is low, I 'tricked' the job into thinking it was uploading a complete.mar but in fact it was just a small checksums file. Hence why https://public-artifacts.taskcluster.net/YkE_OQjoTj2G92P4A4Cy1A/0/public/manifest.json is such a small filesize
  2. the bucket is not archive.m.o, I used a temp one created here: https://console.aws.amazon.com/s3/home?region=us-east-1#&bucket=mozilla-releng-beetmoverscript&prefix=pub/firefox/nightly/2017/01/2017-01-06-16-03-57-date/

sorry about the large'ish patch. I wanted to break it into contained commits but changes ended being a domino effect and by the end, everything was touched. :)

@MihaiTabara @escapewindow 🗿 📜 ✂️  for who has to r? :)


escapewindow:
The test failures seem to be missing files... once I added a dummy `test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/target.txt`, it started complaining about a missing `test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/target.mozinfo.json`.

Git claims that `target.txt` was blocked by the .gitignore, which may be the `build/` rule which is meant to ignore the directory `python setup.py sdist` creates.  Potential fixes: we can figure out how to fix the rule, or `git add -f`, or extract a tarball with the artifacts in it.  `git add -f` is probably the fastest path to victory.

README.md
```diff
@@ -66,22 +66,22 @@ artifact_dir: "/app/beetmoverworker/artifact"
...
+base_gpg_home_dir: "/app/beetmoverworker-1.0.0b4/gpg"                                                                                 [48/9992]
+gpg_lockfile: "/app/beetmoverworker-1.0.0b4/gpg_homedir.lock"
+git_key_repo_dir: "/app/beetmoverworker-1.0.0b4/key_repo"
+git_commit_signing_pubkey_dir: "/app/beetmoverworker-1.0.0b4/valid_git_fingerprints/"
+last_good_git_revision_file: "/app/beetmoverworker-1.0.0b4/git_revision"
+pubkey_path: "/app/beetmoverworker-1.0.0b4/my_pubkey.asc"
+privkey_path: "/app/beetmoverworker-1.0.0b4/my_privkey.asc"
 gpg_path: /usr/local/bin/gpg
-my_email: "scriptworker@example.com"
+my_email: "cottester@gmail.com"
```
escapewindow: Is this a real address?  example.com is a standard dummy domain name; gmail is a real domain, so cottester@ may end up sending mail to a real email.

Beyond that, `my_email` is used to determine which gpg key to use in scriptworker, so this will probably require adding a `cottester@gmail.com` gpg key.  Unless we have a reason to change this, I think we should keep it.

README.md
```diff
@@ -66,22 +66,22 @@ artifact_dir: "/app/beetmoverworker/artifact"
...
-last_good_git_revision_file: "/app/beetmoverworker/git_revision"
-pubkey_path: "/app/beetmoverworker/my_pubkey.asc"
-privkey_path: "/app/beetmoverworker/my_privkey.asc"
+base_gpg_home_dir: "/app/beetmoverworker-1.0.0b4/gpg"                                                                                 [48/9992]
+gpg_lockfile: "/app/beetmoverworker-1.0.0b4/gpg_homedir.lock"
+git_key_repo_dir: "/app/beetmoverworker-1.0.0b4/key_repo"
+git_commit_signing_pubkey_dir: "/app/beetmoverworker-1.0.0b4/valid_git_fingerprints/"
+last_good_git_revision_file: "/app/beetmoverworker-1.0.0b4/git_revision"
+pubkey_path: "/app/beetmoverworker-1.0.0b4/my_pubkey.asc"
+privkey_path: "/app/beetmoverworker-1.0.0b4/my_privkey.asc"
```
escapewindow: Do we want to keep these dir suffix changes?  If so, a) this doesn't seem to match the other `/app/beetmoverworker` config items, and b) we're about to release a scriptworker 1.0.0b5; do we want to have to update this readme every time we do a scriptworker release?

README.md
```diff
@@ -122,23 +122,42 @@ my_email: "scriptworker@example.com"
...
+    "upload_date": 1483465253,
+    "update_manifest": true,
+    "upstreamArtifacts": [
+      {
+        "paths": [
+          "public/build/balrog_props.json"
+        ],
+        "taskId": "c6DPRNsTRa6uVYVp5qCD9w",
+        "taskType": "build",
+        "locale": "en-US"
```
escapewindow: Wondering if we will ever have different locale lists between the balrog_props and the other artifacts.  If so, having locales specified per `upstreamArtifact` dict makes sense.  If not, specifying them outside might make sense.

beetmoverscript/script.py
```diff
@@ -12,16 +12,16 @@
 import aiohttp
 import boto3
 
-from scriptworker.client import get_task, validate_artifact_url
+from scriptworker.client import get_task
 from scriptworker.context import Context
 from scriptworker.exceptions import ScriptWorkerTaskException, ScriptWorkerRetryException
-from scriptworker.utils import (retry_async, download_file,
-                                raise_future_exceptions, retry_request)
+from scriptworker.utils import retry_async, download_file, raise_future_exceptions
```
escapewindow: I think we don't call `retry_download` anywhere but in the tests anymore, so we can probably remove it.

README.md
escapewindow: We may also be able to remove the `valid_artifact_rules` references.

beetmoverscript/templates/fennec_nightly_en_us_multi_signed.yml
```diff
@@ -6,12 +6,10 @@ mapping:
...
   {% set update_balrog_manifest = True if locale == 'multi' else False %}
-    apk:
-      artifact: {{ locale_prefix }}target.apk
+    target.apk:
       s3_key: {{ locale_prefix  }}fennec-{{ version }}.{{ locale }}.android-arm.apk
       update_balrog_manifest: {{ update_balrog_manifest }}
     # Bug 1316124 - TODO
-    #checksums:
-      #artifact: {{ locale_prefix }}target.checksums
+    #target.checksums:
```
MihaiTabara: Not related to this change, but since we're here, are the Fennec signing tasks going to upload the checksums too?

beetmoverscript/templates/fennec_nightly_repacks_signed.yml
```diff
@@ -3,15 +3,11 @@
...
-    apk:
-      artifact: target.apk
+    target.apk:
       s3_key: fennec-{{ version }}.{{ locale }}.android-arm.apk
       update_balrog_manifest: True
-    # TODO: to add checksums once we enable COT
-    #checksums:
-      #artifact: target.checksums
+    # Bug 1316124 - figure out checksums
+    #target.checksums:
```
MihaiTabara: As above, unrelated to the code change, but a reminder for us all to clarify this before Tier-1 (bug 1316124)

beetmoverscript/templates/fennecx86_nightly_en_us_multi_signed.yml
```diff
@@ -6,12 +6,10 @@ mapping:
...
-    apk:
-      artifact: {{ locale_prefix }}target.apk
+    target.apk:
       s3_key: {{ locale_prefix  }}fennec-{{ version }}.{{ locale }}.android-i386.apk
       update_balrog_manifest: {{ update_balrog_manifest }}
-    # Bug 1316124 - TODO
-    #checksums:
-      #artifact: {{ locale_prefix }}target.checksums
+    # Bug 1316124 - TODO figure out checksums
+    #target.checksums:
```
MihaiTabara: As above, unrelated to the code change, but a reminder for us all to clarify this before Tier-1 (bug 1316124)

README.md
lundjordan: yeah, this was added late to the commit and I forgot to strip the 1.0.0b4 from the paths. I have a few beetmover versions locally. Will fix.

README.md
lundjordan: again, similar to the previous beetmoverscript-1.0.0b4 path issue

README.md
lundjordan: I'm not sure. I can't remember if repacks use the Nightly build (which is en-US) balrog_props.json. I think I'll leave it at the taskid level for now. This also allows us to do more than one repack in a task if we ever want to do so.

beetmoverscript/script.py
lundjordan: thanks. will remove

lundjordan:
thanks for the scan aki!

yeah, tests worked locally because those files existed but I didn't notice they were gitignored when I was adding everything to the commit. I'll address this too

beetmoverscript/templates/firefox_nightly_en_us_signed.yml
```diff
@@ -5,20 +5,16 @@ mapping:
...
       s3_key: firefox-{{ version }}.{{ locale }}.{{ platform }}.complete.mar
       update_balrog_manifest: true
-    checksums:
-      artifact: target.checksums
-      s3_key: firefox-{{ version }}.{{ locale }}.{{ platform }}.checksums
-    checksums_asc:
-      artifact: target.checksums.asc
-      s3_key: firefox-{{ version }}.{{ locale }}.{{ platform }}.checksums.asc
-
+# Bug 1316124 - figure out checksums
```
MihaiTabara: I think for Firefox it's safe to uncomment these. The files are correctly attached to the signing task and are specified in the upstreamArtifacts. At least in your fake beetmover task https://queue.taskcluster.net/v1/task/YkE_OQjoTj2G92P4A4Cy1A as well as in the README. IIUC we ought to tweak the firefox/fennec in-tree tasks as well here https://bug1317800.bmoattachments.org/attachment.cgi?id=8823538 to encompass the checkums?

beetmoverscript/templates/firefox_nightly_en_us_unsigned.yml
```diff
@@ -4,70 +4,50 @@ mapping:
...
       s3_key: mozharness.zip
-    xpi:
-      artifact: {{ platform }}/xpi/firefox-{{ version }}.{{ locale }}.langpack.xpi
-      s3_key: firefox-{{ version }}.{{ locale }}.langpack.xpi
-    mar_tools_mar:
-      artifact: host/bin/mar
+    # uncomment when `Bug 1329310 - Support SIMPLE NAME for langpacks too` is resolved
+    #    target.langpack.xpi:
+    #      s3_key: firefox-{{ version }}.{{ locale }}.langpack.xpi
+    mar:
```
MihaiTabara: Just a heads-up on `mar` and `mbsdiff`. Haven't tested yet but are we 100% sure that the files are downloaded in the `work_dir/cot/{task_id}/{mar,mbsdiff}` accordingly? Just worried a bit of that "/host/bin" prefix in the non-CoT world.

beetmoverscript/templates/firefox_nightly_en_us_unsigned.yml
```diff
@@ -4,70 +4,50 @@ mapping:
...
 
   {% if platform in ["linux-i686", "linux-x86_64"] %}
-    sdk:
-      artifact: target.sdk.tar.bz2
+    target.sdk.tar.bz2:
       s3_key: firefox-{{ version }}.{{ locale }}.{{ platform }}.sdk.tar.bz2
     # Bug 1316125 - TODO
-    #sdk_asc:
-      #artifact: target.sdk.tar.bz2.asc
+    #target.sdk.tar.bz2.asc:
```
MihaiTabara: Hm, good point. We need to double check with @nthomas-mozilla on this before Tier-1.

beetmoverscript/test/__init__.py
```diff
@@ -10,7 +10,7 @@ def get_fake_valid_task():
 
 
 def get_fake_balrog_props():
-    return load_json(path="beetmoverscript/test/fake_balrog_props.json")
```
MihaiTabara: I think we can safely remove `fake_balrog_manifest.json` now that we have `test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/balrog_props.json`

beetmoverscript/test/test_script.py
```diff
@@ -100,17 +103,27 @@ def test_move_beets(event_loop):
...
     context.properties = get_fake_balrog_props()["properties"]
     context.properties['platform'] = context.properties['stage_platform']
-    manifest = generate_candidates_manifest(context)
+    context.artifacts_to_beetmove = get_upstream_artifacts(context)
+    manifest = generate_beetmover_manifest(context.config, context.task, context.properties)
 
     expected_sources = [
-        'https://queue.taskcluster.net/v1/task/VALID_TASK_ID/artifacts/public/build/target.package',
-        'https://queue.taskcluster.net/v1/task/VALID_TASK_ID/artifacts/public/build/en-US/target.package'
+        os.path.abspath(
```
MihaiTabara: Need to actually create/touch dummy files for {target.mozinfo.json, target.txt, target_info.txt} under test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build in order for tests to pass

MihaiTabara:
@lundjordan Sorry about overlapping comments with @escapewindow's scan. Dove in the review-mode and forgot to refresh/double-check with previous comments before pushing the review.

beetmoverscript/script.py
```diff
@@ -30,53 +30,37 @@
 async def async_main(context):
     # balrog_manifest is used by a subsequent balrogworker task that points to a beetmoved artifact
```
lundjordan: sure. but 'this beetmover artifact' might be ambiguos since there are many artifacts. I'll clean up my comment though.

beetmoverscript/templates/fennec_nightly_en_us_multi_signed.yml
lundjordan: yeah, I wasn't sure about checksums at all. will probably have to discuss @ tues mtg

beetmoverscript/templates/fennecx86_nightly_en_us_multi_signed.yml
lundjordan: should add this too to the mtg

beetmoverscript/templates/firefox_nightly_en_us_unsigned.yml
lundjordan: yes, while it seems like this code would mean the mar/mbsdiff artifacts would live in `work_dir/cot/{task_id}/{mar,mbsdiff}`, they actually don't. They are still at `work_dir/cot/{task_id}//host/bin/{mar,mbsdiff}` or whatever. This is because 'mar' and 'mbsdiff' is just the key in context.artifacts_to_beetmove[locale] dict that maps artifact basename to full abs_paths.

see:

- https://github.com/mozilla-releng/beetmoverscript/blob/cot/beetmoverscript/task.py#L48
- https://github.com/mozilla-releng/beetmoverscript/blob/cot/beetmoverscript/script.py#L58-L60

Does this clear things up or did I miss something in my logic?

beetmoverscript/templates/fennecx86_nightly_en_us_multi_signed.yml
MihaiTabara: sounds good, done

beetmoverscript/test/__init__.py
lundjordan: ah, good catch 👍

beetmoverscript/test/test_script.py
lundjordan: thanks. fixed via https://github.com/mozilla-releng/beetmoverscript/pull/26/commits/f781d93d525005eb676eabb7fed7be851289f811

beetmoverscript/script.py
```diff
@@ -12,16 +12,16 @@
 import aiohttp
```
lundjordan: out of the three current scriptworker implementations, balrogscript is the odd one out:

https://github.com/mozilla-releng/beetmoverscript/tree/master/beetmoverscript
https://github.com/mozilla-releng/signingscript/tree/master/signingscript
https://github.com/mozilla-releng/balrogscript/tree/master/balrogscript

I agree consistency would be better. Happy to go with {scriptworker_type}script.py naming convention if that's the way we want it.

beetmoverscript/constants.py
```diff
@@ -11,13 +13,17 @@
...
     '.apk': 'application/vnd.android.package-archive',
 }
-
-PLATFORM_MAP = {
+STAGE_PLATFORM_MAP = {
     'linux': 'linux-i686',
     'linux64': 'linux-x86_64',
 }
-
+# TODO don't rely on default dict
```
lundjordan: sort of. This code copies this logic:
https://github.com/mozilla-releng/beetmoverscript/blob/47aa586ad10632a178991a01dad8e1335ae2b017/beetmoverscript/utils.py#L58-L62

I could put something like `'linux': 'firefox'`  lines in here. But that would require me to sanity check what the linux 32 and 64 names would be and all the other osx/win platforms. So I was sort of lazy here. :)

I figured we could be more explicit when we port osx/win over.

beetmoverscript/templates/firefox_nightly_en_us_signed.yml
lundjordan: like the other places, I'll leave all signing task based checksum files out of the manifest until we discuss

lundjordan:
@MihaiTabara final r? with last two in-pr commits?

beetmoverscript/script.py
MihaiTabara: Bleah, apologies for being lazy and not check. I guess it makes sense though "signing-script", "balrog-script", "beetmover-script". I'll add a new issue in balrogscript to correct it and align it consistenly with others, pending the follow-up packaging.

beetmoverscript/constants.py
MihaiTabara: 👍 TIL lamda can default to some value  :)

beetmoverscript/script.py
```diff
@@ -28,7 +28,9 @@
 
 # async_main {{{1
 async def async_main(context):
-    # balrog_manifest is used by a subsequent balrogworker task that points to a beetmoved artifact
+    # balrog_manifest is written and uploaded as an artifact which is used by a subsequent
```
MihaiTabara: awesome comment \o/ 👍

PR #27, opened 2017-01-11
Bug 1317747- Add scopes check in beetmoverscript

MihaiTabara:


beetmoverscript/task.py
```diff
@@ -17,6 +19,19 @@ def validate_task_schema(context):
     scriptworker.client.validate_json_schema(context.task, task_schema)
 
 
+def validate_task_scopes(context):
+    scopes = context.task['scopes']
+    for scope in scopes:
```
escapewindow: This tests whether the scope exists.  I would like us to also only allow pushing to a nightly bucket / use nightly creds if we are using `project:releng:beetmover:nightly`, and same for release bucket / release creds if we're using `project:releng:beetmover:release`.

MihaiTabara: Addressed in follow-up

beetmoverscript/task.py
```diff
@@ -17,6 +19,27 @@ def validate_task_schema(context):
...
+                break
+            log.warning('scope {} is malformed, skipping!'.format(scope))
+    else:
+        log.critical("No beetmover scopes!")
+        sys.exit(3)
+
+    # prevent uncoordination between the bucket supposed to use for beetmove
+    # and the current credentials scopes
+    if (signing_cert_name not in manifest['s3_prefix_dated'] or
+            signing_cert_name not in manifest['s3_prefix_latest']):
```
escapewindow: Thanks for adding this!

Now I'm worried about someone pushing to `pub/mobile/nightly/...` where `...` contains `dep` in the string, but isn't part of the path.

I'm not sure how hardcoded we want to go, but we could do something like

```python
restricted = {
    'pub/mobile/nightly': 'nightly',
    'pub/firefox/nightly': 'nightly',
    'pub/mobile/release': 'release',
    'pub/firefox/release': 'release',
}

for k, v in restricted:
    for path in (manifest['s3_prefix_dated'], manifest['s3_prefix_latest']):
        if path.startswith(k) and scope != v:
            # fatal error
```

Approved with that change :)
I'll go ahead and file a separate creds issue that we can do later.

MihaiTabara: Yeah, this makes perfect sense! If someone would have CI scopes like "project:releng:beetmover:try" and manifests like 's3_prefix_dated' = "pub/mobile/nightly/2017/01/2017-01-try.." and/or 's3_prefix_latest' = "pub/mobile/nightly/2017/01/2017-01-try.." it would move on to pushing to nightly without barfing ...

MihaiTabara: I am not sure if the existing check was to be replaced by this or just enriched. For safety reasons I left it too. I am thinking that any scopes munging to access the nightly/release buckets will be prevented by your suggested check. The existing one should catch munging scopes across various low-security trees trying to access each other buckets. Not sure if it makes sense. Happy to remove it if it doesn't. 

I'll go ahead and land this for now and deployed it on workers. Will need to make changes to address the checksums anyway so we'll most likely follow-up with future bumps anyway.

escapewindow: I would have leaned towards removing the previous check.  One side effect of leaving the previous check is @lundjordan won't be able to test until he changes the in-tree scopes to `project:releng:beetmover:nightly`, since the `s3_prefix`es don't contain `dep`.

At some point, maybe let's change the `s3_prefix`es to be "pub/mobile/{scope}/..." and format.
#28  will also help here.

PR #31, opened 2017-01-13
adds aiohttp_max_connections to limit concurrency in script run

lundjordan:


lundjordan:
needs aiohttp_max_connections added to the script's config in puppet. I believe beetmoverscript nodes are currently pinned to https://github.com/MihaiTabara/build-puppet/tree/cot

PR #32, opened 2017-01-18
Bug 1305139 - Remove all current checksums. Need to reimplement the m…

MihaiTabara:
…echanism.


beetmoverscript/task.py
```diff
@@ -43,6 +43,13 @@ def validate_task_scopes(context, manifest):
                 sys.exit(3)
 
 
+def add_checksums_manifest_to_artifacts(context, filename):
+    abs_file_path = os.path.join(context.config['artifact_dir'],
+                                 'public/{}'.format(filename))
```
escapewindow: We can put this in `public/`, or a subdir of `public/`.  I'm not sure which makes more sense.  This works.

beetmoverscript/script.py
```diff
@@ -86,6 +103,20 @@
...
         )
 
+    context.checksums_manifest.extend(enrich_checksums_manifest(source, pretty_name))
+
+
+def enrich_checksums_manifest(path, pretty_name):
+    digest_info = []
+
+    size = get_size(path)
+    for algo in ("sha512", "md5", "sha1"):
```
escapewindow: We may be calculating the same hash multiple times (`get_hash` here and line 115).
If we want to speed things up, we may want to create a dictionary with the various name, size, and hash info, and then populate templates with that information.

escapewindow: We seem to be calling `get_size` multiple times as well, although that may be faster than `get_hash`.

beetmoverscript/script.py
```diff
@@ -33,26 +35,38 @@
...
     #   a. map each upstream artifact to pretty name release bucket format
     #   b. upload to candidates/dated location
     await move_beets(context, context.artifacts_to_beetmove, mapping_manifest)
-    # # 6. write balrog_manifest to a file and add it to list of artifacts
+    #  write balrog_manifest to a file and add it to list of artifacts
     if context.task["payload"]["update_manifest"]:
         add_balrog_manifest_to_artifacts(context)
+    # determine the correct checksum filename and generate it, adding it to
+    # the list of artifacts afterwards
+    if context.task["payload"]["generate_checksums"]:
```
lundjordan: aki mentioned that we may not need this boolean here: https://bugzilla.mozilla.org/show_bug.cgi?id=1305139#c7

MihaiTabara:
@escapewindow @lundjordan 
Thanks for feedback. I've addressed it and refactored. Few things:
1. The "generate_checksums" is indeed useless at this point. However, I'm keeping still for another few hours till I manage to fix the tests failing in its absence. 
2. I still need to double check the logic for l10n
3. I need to double-check what happens with the second BM job that grabs the signed checksum file + detached signature. The existing logic will attempt to pretty name it and will fail. I need to revisit the templates to potentially add the checksums + signature files there with a small hack (s3_key be both the key and the s3key).

beetmoverscript/script.py
```diff
@@ -61,33 +79,41 @@
...
 
-async def move_beet(context, source, destinations, locale, update_balrog_manifest):
-    beet_config = deepcopy(context.config)
-    beet_config.setdefault('valid_artifact_task_ids', context.task['dependencies'])
-
+async def move_beet(context, source, destinations, locale,
+                    update_balrog_manifest, pretty_name):
     await retry_upload(context=context, destinations=destinations, path=source)
 
+    context.checksums_dict.setdefault(pretty_name, {
```
escapewindow: I think you're using `setdefault` to avoid setting the checksum for a file multiple times.  If you didn't care about that, you could have

```python
context.checksums_dict[pretty_name] = {algo: get_hash(source, algo) for algo in CHECKSUMS_DIGESTS}
```

instead.

However, the right side of `setdefault()` gets evaluated either way.  For instance,

```python
#!/usr/bin/env python
from __future__ import print_function
d = {'a': 1}
d.setdefault('a', print("foo"))  # This shouldn't set d['a'], because it's already set
print(d)
```

The output from the above scriptlet is

```
foo
{'a': 1}
```

So the dictionary `d` isn't modified because we used `setdefault`, and `d['a']` is already set.  However, we evaluate the `print(foo)` anyway.  That's pretty harmless here, but this means is we'll run `get_hash` for all the `CHECKSUMS_DIGESTS`, even if `contest.checksums_dict[pretty_name]` is already defined, wasting cycles.

Instead, (if we are worried about running `get_hash` for the same file multiple times), we could

```python
if context.checksums_dict.get(pretty_name) is None:
    context.checksums_dict[pretty_name] = {algo: get_hash(source, algo) for algo in CHECKSUMS_DIGESTS}
```

I think that'll do what you want.

beetmoverscript/task.py
```diff
@@ -43,6 +44,28 @@ def validate_task_scopes(context, manifest):
                 sys.exit(3)
 
 
+def dump_checksums_manifest(checksums_dict):
+    manifest = list()
+    for artifact, values in checksums_dict.items():
```
escapewindow: If you wanted this alphabetically sorted by filename, you could

```python
    for artifact, values in sorted(checksums_dict.items()):
```

beetmoverscript/task.py
```diff
@@ -43,6 +43,29 @@ def validate_task_scopes(context, manifest):
...
+                values['size'],
+                artifact
+            ))
+
+    return manifest
+
+
+def add_checksums_manifest_to_artifacts(context, filename):
+    abs_file_path = os.path.join(context.config['artifact_dir'],
+                                 'public/{}'.format(filename))
```
escapewindow: I'm fine leaving it where it is. Thanks @MihaiTabara !

MihaiTabara:
@lundjordan @escapewindow 
Okay, so I have wrapped up the beetmoverscript side to make it as generic as possible for future use. I addressed both your and mine previous comments. 

Observations:
* from now one, *all* beetmover tasks will generate, regardless, a `target.checksums` file containing the digests for files that it beetmoves to S3

In our particular case of nightlies, generically we'd need to schedule:
1. BM1 - a task that depends on both `build` and `signing` tasks (to gather all artifacts that are needed to constitute the checksums file)

2. a SBM1 - a signing task consuming the `target.checksums` file and returning a signed `target.checksums` + `target.checksums.asc`

3. a follow-up BM2 depending on SBM1, consuming the checksums file + its detached signature, pretty names them and then (naturally) uploads them to S3.

Bonus:
* For Fennec l10n jobs we don't need BM1 as that's already the existing l10n signing tasks.

escapewindow:
I think so.  Don't forget to add the checksums to the puppet `script_config.json`!

MihaiTabara:
Yep. And also tweak the new templates under https://hg.mozilla.org/build/puppet/file/tip/modules/beetmover_scriptworker/templates/script_config.json.erb#l12

PR #34, opened 2017-01-25
Bug 1333826 - Stop shipping the Firefox SDK on all platforms

bsmedberg:
r? @lundjordan

### DIFF
--- a/beetmoverscript/templates/firefox_nightly_en_us_unsigned.yml
+++ b/beetmoverscript/templates/firefox_nightly_en_us_unsigned.yml
@@ -40,13 +40,4 @@ mapping:
       s3_key: mar-tools/{{ stage_platform }}/mar
     mbsdiff:
       s3_key: mar-tools/{{ stage_platform }}/mbsdiff
-
-  {% if platform in ["linux-i686", "linux-x86_64"] %}
-    target.sdk.tar.bz2:
-      s3_key: firefox-{{ version }}.{{ locale }}.{{ platform }}.sdk.tar.bz2
-    # Bug 1316125 - TODO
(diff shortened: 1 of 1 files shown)


lundjordan:
@bsmedberg fyi, we will have to remove the following lines on both mozilla-central and mozilla-aurora prior to merging this patch: https://dxr.mozilla.org/mozilla-central/source/taskcluster/taskgraph/transforms/beetmover.py#94,97

It's unfortunate for now but mach taskgraph is the source of truth for files, this patch here is simply the mapping dict. In the future we hope to only have to define the list of upload artifacts in one place, at the build sys level: https://bugzilla.mozilla.org/show_bug.cgi?id=1331143

bsmedberg:
Yes, that's the patch at https://bugzilla.mozilla.org/attachment.cgi?id=8830460

MihaiTabara:
@lundjordan @bsmedberg 
If this is not urgent, I can take this change and add it under https://github.com/mozilla-releng/beetmoverscript/pull/32/ which will make a larger change with respect to generating checksums.

MihaiTabara:
@lundjordan @bsmedberg
I've included your change in the PR I'm about to land. Please double check here - https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/templates/firefox_nightly.yml.

Feel free to close the PR after that. Thanks and sorry for hijacking the change! ^_^

MihaiTabara:
Nevermind me, I added back the SDK files as I didn't want to land the other patches to aurora + nightly without r?.

MihaiTabara:
Did a manual merge. Thanks for patching this!

PR #38, opened 2017-01-30
bug 1333895 - reduce max concurrent connections

escapewindow:
This is good for new installs; we'll also need a puppet patch for prod installs.


escapewindow:
Beetmoverscript's test died because of https://github.com/taskcluster/taskcluster-client.py/issues/58 .  We can pin taskcluster<1.0.0 like https://github.com/escapewindow/scriptworker/commit/ec846f8cda15bb5b4f99b97099e403ceea641146 , or wait until a fixed version is released.

MihaiTabara:
I've retriggered the build to pickup 1.0.2 as per https://github.com/taskcluster/taskcluster-client.py/issues/58#issuecomment-276344180 and it looks good.

PR #39, opened 2017-02-01
add path to upload log status

escapewindow:
When we upload many artifacts in parallel, we often see many log lines with a status line but no sign of which artifact was uploaded.  By adding the path to the log, we'll be able to trace this easier.


MihaiTabara:
Thanks 👍

PR #42, opened 2017-02-03
Bug 1305139 - add balrog_props.json to generated artifacts. r=aki

MihaiTabara:


escapewindow:
15:14 <aki> hm, we don't download the file?
we can
```
release_props_file = get_initial_release_props_file(context)
context.release_props = get_release_props(release_props_file)
# copy release_props_file to artifact_dir
```
15:17 <aki> probably a shutil.copyfile

beetmoverscript/task.py
```diff
@@ -58,6 +59,14 @@ def generate_checksums_manifest(context):
     return '\n'.join(content)
 
 
+def regenerate_release_props(context):
```
escapewindow: we don't need this function anymore.

PR #46, opened 2017-02-07
Bug 1336926 - fix broken test-packages.json. r=aki

MihaiTabara:


beetmoverscript/utils.py
```diff
@@ -119,3 +119,25 @@ def get_release_props(initial_release_props_file, platform_mapping=STAGE_PLATFOR
...
+def alter_unpretty_contents(original_artifact_name, source, artifact_pretty_name):
+    """
+    Function to alter any unpretty-name contents from a file.
+    Args:
+        original_artifact_name: e.g. `target.test_packages.json`
+        artifact_pretty_name: e.g. `firefox-53.0a1.en-US.linux-i686.test_packages.json`
+        source: path to CoT file
+    """
+    # remove `target` keyword and obtain the filename
+    filename = original_artifact_name.split('.', 1)[1]
```
escapewindow: We can take every simple name and pass it through the manifest, right?  (We may have to use `os.path.basename` to avoid using the `en-US/` subdir)  I'm wondering if that may be a better way.

Your method works, as long as the `target` name always matches the pretty name (which might not be true).  We already have a way to translate the simple to the pretty name, though.

beetmoverscript/test/test_work_dir/task.json
```diff
@@ -19,6 +19,7 @@
             "public/build/target.txt",
             "public/build/target.mozinfo.json",
             "public/build/target_info.txt",
+            "public/build/target.test_packages.json",
```
escapewindow: It looks like we have to populate `test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/target.test_packages.json` for `test_move_beets` to work.

escapewindow:
Thanks for fixing the tests!
Are we still planning on using the manifest to update the json?

MihaiTabara:
@escapewindow Yes, definitely. Thanks for suggesting that again. Just resolving the issues one by one :)

MihaiTabara:
@escapewindow 

So thankful you've requested the switch to reusing mapping manifests. Found two bugs along the way:
1. We missed `crashreporter-symbols full` and `gtests` in the firefox nightly mapping manifest. Even though we don't ship them currently in neither of the tasks, they are still needed in repolishing the `test_packages.json` file

2. Reusing the mapping manifests `s3_keys` automatically solved the `jssshell zip` thing which is a particular case in the artifacts. Even though it is originally called `target.jsshell.zip` it doesn't get translated to pretty naming as other, but differently to `jsshell-linux-i686.zip` (so missing all the product, version, etc).

beetmoverscript/script.py
```diff
@@ -56,6 +57,12 @@
     # validate scopes to prevent beetmoving in the wrong place
     validate_task_scopes(context, mapping_manifest)
 
+    # some files to-be-determined via script configs need to have their
+    # contents pretty named, so doing it here before even beetmoving begins
+    blobs = context.config.get('blobs_needing_prettynaming_contents', [])
+    if blobs:
```
escapewindow: This check actually isn't needed.  If `blobs == []`, then the `for` loop in `alter_unpretty_constants` will be noop.

MihaiTabara: Ah, good catch! 
I'll refactor and merge 👍

PR #48, opened 2017-02-09
Finally send update_manifest to graveyard.r=aki

MihaiTabara:


MihaiTabara:
@escapewindow 
Now that checksums in-tree tasks generation is landed on `central` and we've gotten `0.2.2` beetmoverscript up-to-date with all changes, we can retire `update_manifest` completely. The change will get picked-up with the next beetmoverscript bump, whenever that'll be in the next days/week.

PR #49, opened 2017-02-13
Bug 1338186 - make beetmoverscript work for Fennec.

MihaiTabara:
@escapewindow 

This thing is driving me nuts. I think I've been too greedy to make it super-generic from the first iteration. I've changed so much in the logic that I've lost my way in the code. I could really use some help. Initially, I wanted to push my whole set of patches but I realized everything depends on the way the `script_config` is built. Please let me know if this version of script_config reflects your opinion from last week. If yes, then I'll work to unravel the things. If no, please make comments so that I can address accordingly. 

Thank you!


config_example.json
```diff
@@ -15,19 +15,146 @@
...
         "firefox": {
-            "bucket": "mozilla-releng-beetmoverscript-firefox",
-            "credentials": {
-                "id": "dummy",
-                "key": "dummy"
+            "nightly": {
+                "push-to-nightly": {
+                    "s3_prefixes": {
+                        "default": {
+                            "s3_prefix": "pub/firefox/nightly/{}-{}/",
```
escapewindow: Using `"{}".format(var)` is handy, but it means we have to know order.

We can also use format with kwargs, or % strings with kwargs, which allows us to replace without having to know or care about order:
```
>>> "{}".format("a")
'a'
>>> "{two}{one}".format(one=1, two=2)
'21'
>>> "%(two)s %(one)s" % {"one": 1, "two": 2}
'2 1'
```

That also allows us to have a generic dictionary to replace things with.  For example, 

```
>>> "{two}{one}".format(one=1, two=2, three=3)
'21'
>>> "%(two)s %(one)s" % {"one": 1, "two": 2, "three": 3}
'2 1'
```
still work even though `three` isn't in the template.

config_example.json
```diff
@@ -15,19 +15,146 @@
...
+                            "s3_prefix_latest": "pub/firefox/nightly/latest-{}/"
+                        },
+                        "repacks": {
+                            "s3_prefix": "pub/firefox/nightly/{}-{}-l10n/",
+                            "s3_prefix_latest": "pub/firefox/nightly/latest-{}-l10n/"
+                        }
+                    },
+                    "bucket": "mozilla-releng-firefox-nightly-bucket",
```
escapewindow: So this is a bit different.

We were discussing having the bucket/creds be one thing (nightly vs release vs dep scopes), and behavior be another thing (push-to-nightly vs push-to-candidates etc).  This latter thing will be determined by another item in the task definition, like `task.payload.behavior` or `task.payload.beetmover.action` or something.  This could also be a second scope, similar to how signingscript allows for a [single cert type](https://github.com/mozilla-releng/signingscript/blob/master/signingscript/task.py#L25) (nightly vs dep vs release) but allows for signing formats to be specified in [another set of scopes](https://github.com/mozilla-releng/signingscript/blob/master/signingscript/task.py#L34) (gpg vs jar vs mar)

So if I were you, I'd move all the bucket/credential stuff to their own area, e.g.

```
"buckets": {
    "nightly": {  # based on project:releng:beetmover:nightly scopes
        "bucket": "...",
        "credentials": {...},
    },
    "release": {...},
    "dep": {...},
}
```

and then specify the actions / behaviors etc in their own dict, e.g.

```
"actions": {
    "push-to-nightly": {
        # path templates
     },
    "push-to-candidates": {
        # path templates
    },
```
These may someday get more than just path templates, so I'd leave as a dict rather than hardcode to a list, but up to you.  Also, depending on the action(s) in the task definition, I'd call different functions:

```
action_map = {
    'push-to-nightly': push_to_nightly_function,
    'push-to-candidates': push_to_candidates_function,
}
    ...
    context.task = get_task(context.config)
    context.action = get_action(context.task)
    if action_map.get(context.action):
        await action_map[context.action](context)
    else:
        log.critical("Unknown action {}!".format(context.action))
        # die
```

Does this make sense?  This is what I was saying on the whiteboard, but I'm not sure if I communicated that clearly.

config_example.json
MihaiTabara: You're right. I should've sticked to these. I started using them initially but then turned to `format` for whatever reasons. Makes sense, thanks!

config_example.json
MihaiTabara: Apologies for misunderstanding. My bad - it all makes sense now.
I've stashed my work and started fresh. Thanks for clarifications!

MihaiTabara:
Closing this as development is now tracked in a separate PR https://github.com/mozilla-releng/beetmoverscript/pull/51

PR #51, opened 2017-02-16
Bug 1338186 - add beetmoverscript support for releases

MihaiTabara:


beetmoverscript/utils.py
```diff
@@ -51,12 +51,12 @@ def write_file(path, contents):
         fh.write(contents)
 
 
-def generate_beetmover_template_args(task, release_props):
+def generate_beetmover_template_args(task, release_props, cert, action):
     tmpl_key_platform = TEMPLATE_KEY_PLATFORMS[release_props["stage_platform"]]
```
MihaiTabara: This `TEMPLATE_KEY_PLATFORMS` is solely used here to toggle between the template files (`fennec_nightly` vs `fennecx86_nightly`). But looking at the two templates for the two different Android platforms, I noticed they only differ by "arm" vs "i386 in labels. I could easily merge them and just toggle that part via a template arg. That way we get rid both of this env dict, the `fennecx86_*` yml files and the logic from this function. What do you think?

beetmoverscript/utils.py
```diff
@@ -65,22 +65,28 @@ def generate_beetmover_template_args(task, release_props):
         "platform": release_props["platform"],
     }
 
+    if cert != 'nightly':
+        template_args["build_number"] = release_props["buildnumber"]
```
MihaiTabara: I meant to ask this before. Will we have the build number coming from somewhere within the `balrog_props` or we need to find another way to pass that?

beetmoverscript/templates/fennec_candidates.yml
```diff
@@ -0,0 +1,114 @@
...
+        - {{ platform }}/{{ locale }}/fennec-{{ version }}.{{ locale }}.android-arm.txt
+    target.web-platform.tests.zip:
+      s3_key: fennec-{{ version }}.{{ locale }}.android-arm.web-platform.tests.zip
+      destinations:
+        - {{ platform }}/{{ locale }}/fennec-{{ version }}.{{ locale }}.android-arm.web-platform.tests.zip
+    target.xpcshell.tests.zip:
+      s3_key: fennec-{{ version }}.{{ locale }}.android-arm.xpcshell.tests.zip
+      destinations:
+        - {{ platform }}/{{ locale }}/fennec-{{ version }}.{{ locale }}.android-arm.xpcshell.tests.zip
+    target_info.txt:
```
MihaiTabara: This file is the reason why I tried a different approach for the candidates templates. @escapewindow please let me know if you feel this is better or worse. This *same* file needs to be copied in three diferent places within the candidates bucket with three different namings. So far it's the only one but there could be others in the desktop part. Rather than adding `if file == target_info.yxy do: stuff od` within the code, I preferred to use this approach. What do you think?

beetmoverscript/script.py
```diff
@@ -76,10 +113,64 @@
...
 
 
-async def move_beets(context, artifacts_to_beetmove, manifest):
+async def move_candidates_beets(context, artifacts_to_beetmove, manifest):
+    beets = []
+    for locale in artifacts_to_beetmove:
+        for artifact in artifacts_to_beetmove[locale]:
+            source = artifacts_to_beetmove[locale][artifact]
+            artifact_pretty_name = manifest['mapping'][locale][artifact]['s3_key']
+            destinations = [os.path.join(manifest["s3_prefix_candidates"],
```
MihaiTabara: @escapewindow If you think the `candidates` templates approach (see question below in the template) is better than what we currently have for nightlies (as s3_key + latest and dated prefixes) we could tweak the `nightly` templates as well and reuse the same logic, hence getting rid of this additional `move_candidatess_beets`. The same with the `push-to-nightly` vs `push-to-candidates` which currently only differ by this function call.

beetmoverscript/utils.py
escapewindow: We need to find a way to pass it.  The easiest way might be the task definition, since the decision task knows the build number, and it modifies the beetmover task definition.  If we want it in balrog_props, we need to pass it into the build, and have the build put it into balrog_props, and then beetmover reads it from balrog_props, which is an extra step worth of work.

beetmoverscript/utils.py
escapewindow: That sounds good to me.

beetmoverscript/constants.py
```diff
@@ -28,12 +28,12 @@
 # release buckets don't require a copy of the following artifacts
 IGNORED_UPSTREAM_ARTIFACTS = ["balrog_props.json"]
 
-RESTRICTED_BUCKET_PATHS = {
-    'pub/mobile/nightly': 'nightly',
-    'pub/firefox/nightly': 'nightly',
-    'pub/mobile/release': 'release',
-    'pub/firefox/release': 'release',
+RESTRICTED_ACTIONS = {
```
escapewindow: Hm, the actions aren't restricted; the s3 creds are restricted.

beetmoverscript/script.py
```diff
@@ -76,10 +113,64 @@
     # the checksums file
     add_release_props_to_artifacts(context, release_props_file)
 
+
+async def push_to_releases(context):
+    pass
```
escapewindow: Maybe `raise NotImplementedError(...)` ?

beetmoverscript/script.py
```diff
@@ -76,10 +113,64 @@
...
+
+
+async def push_to_staging(context):
+    pass
```
escapewindow: Here too.

beetmoverscript/script.py
```diff
@@ -76,10 +113,64 @@
...
+    validate_task_schema(context)
+
+    # determine the task action to toggle the right behavior
+    context.cert_name = get_task_cert_type(context.task)
+    context.action = get_task_beetmover_action(context.cert_name, context.task)
+
+    if action_map.get(context.action):
+        await action_map[context.action](context)
+    else:
+        log.critical("Unknow action {}!".format(context.action))
```
escapewindow: [sp] unknown

beetmoverscript/task.py
```diff
@@ -22,26 +21,38 @@ def validate_task_schema(context):
...
-        sys.exit(3)
-
-    # prevent scopes from low-security tree write in a release-type bucket
-    for k, v in RESTRICTED_BUCKET_PATHS.items():
-        for path in (manifest['s3_prefix_dated'], manifest['s3_prefix_latest']):
-            if path.startswith(k) and signing_cert_name != v:
-                log.critical("Munged low-security tree scopes trying to access"
-                             " nightly/release buckets for beetmover")
-                sys.exit(3)
+def get_task_cert_type(task):
```
escapewindow: Ok, I have a few problems with this function.

* there's a lot of cert and signing verbiage here; I'm guessing this is because you copied the signingscript function?  We probably want more beetmover-type verbiage
* we split the credentials/bucket info from the behavior/action so we could control them separately.  That's so the behavior/action can be run on try, as long as we're using the staging bucket.  So the behavior/action scope isn't restricted; the credentials/bucket scope is.  Where do you check that scope?

In a signingscript analogy, the behavior/action is comparable to signing formats, though we're going to restrict the task to one.  We don't care if you're asking to sign with gpg or jar signing, as long as you have the scope to request it.  However, we want to restrict the ability to ask for signing with the nightly or release keys. These map to the nightly and release buckets / credentials for beetmover.

beetmoverscript/task.py
```diff
@@ -22,26 +21,38 @@ def validate_task_schema(context):
...
+    if re.search('^[0-9A-Za-z_-]+$', signing_cert_name) is None:
+        raise ScriptWorkerTaskException("Scope {} is malformed".format(certs[0]))
+
+    if signing_cert_name not in RESTRICTED_ACTIONS.keys():
+        raise ScriptWorkerTaskException("Invalid scope")
+
+    return signing_cert_name
+
+
+def get_task_beetmover_action(cert, task):
```
escapewindow: This is closer to what we want, but a) you're referencing certs again, and b) we don't need to restrict actions, just make sure that there's 1 and it's valid.  We need to add a bucket/credentials scope check.

beetmoverscript/templates/fennec_candidates.yml
escapewindow: I think it would be great to standardize on this format.
We do need to be careful about the test json.  We currently run the simple->pretty name conversion in the test json.  If we have multiple locations, maybe we use the first destination in the list for that conversion?

beetmoverscript/script.py
MihaiTabara: Yep, will do.

beetmoverscript/script.py
MihaiTabara: Yep, will do.

beetmoverscript/script.py
MihaiTabara: 😊

beetmoverscript/task.py
MihaiTabara: Yeah, I striked-through the old scopes validation function and borrowed these both functions from signing script as they were clearer in my opinion. I'll tweak to a more beetmover-type verbiage, thanks for the heads-up.

I'm a little bit confused by the second part. Might be the Friday night so I'll reread this tomorrow morning with fresh eyes. But I'm not getting which scope check you're talking about:

a) hm, let me try this again - you mean that in the future future, you could trigger behavior/action such as "push-go-nightly' and "push-to-candidates" and "push-to-releases" from *try* (hence with "dep" scope),hence allowing developers test a "nightly" graph, a "candidate" graph or a "release" graph with the constraint that all these go under staging-bucket/pub/{firefox,mobile}/{nightly,candidates,releasess}? 

b) the credentials/bucket scope you are asking for - are you talking about bringing back the check I just deleted, the one with: 

`-RESTRICTED_BUCKET_PATHS = {
-    'pub/mobile/nightly': 'nightly',
-    'pub/firefox/nightly': 'nightly',
-    'pub/mobile/release': 'release',
-    'pub/firefox/release': 'release',`

?

beetmoverscript/task.py
MihaiTabara: Same as above, am a little bit confused to the bucket/credentials scope check you're talking about. More details in the other function.

beetmoverscript/script.py
```diff
@@ -18,54 +18,92 @@
...
     # generate beetmover mapping manifest
-    mapping_manifest = generate_beetmover_manifest(context.config,
-                                                   context.task,
-                                                   context.release_props)
-    # validate scopes to prevent beetmoving in the wrong place
-    validate_task_scopes(context, mapping_manifest)
+    mapping_manifest = generate_beetmover_manifest(context)
+
+    # perform another validation check against the bucket path
+    validate_bucket_paths(context.bucket, mapping_manifest['s3_bucket_path'])
```
MihaiTabara: @escapewindow 
1. Until we have a green-go that permissions from `release creds` to push to nightly paths are removed, I say we keep this additional paths-check. Certainly couldn't hurt.
2. The reason this check is missing from `push_to_nightly` is that I haven't yet touched the nightly templates to reflect the *new standard* (of using a separate pretty naming + destinations in the templates). Once we're past the alpha/beeta tesing I'll make the modifications and adapt the nightly templates too so we can merge the two `move_*_beets` functions.

MihaiTabara:
@escapewindow 
Sorry to insist with this long-disputed branch-patch but I've added some more changes:
* implemented the new "standard" for templates in both firefox/fennec nightlies
* reused the same `move_beets` function now that we have consistency between `candidates` and `nightly` main functions
* collapsed the `push_to_candidates` into `push_to_nightly` as they are identical at this point. I can always split if we need to skip/add some more steps in one or another
* double checked the `test_packages.json` generated works smoothly + minor tweak
* fixed/improved tests to increase a bit the coverage and catch some more corner cases
* addressed the nit from last review

Please let me know if there's anything else I could add/change.
If not, I think I'm ready to bump a alpha/beta release out of this branch and start testing in the wild 👍

escapewindow:
🚢 it!

PR #56, opened 2017-03-07
Bug 1343649 - fix build_number and version within various action types. r=jlorenzo

MihaiTabara:
@escapewindow @JohanLorenzo 

Latest jamun-based staging Fennec beetmover tasks fail because of `jinja2.exceptions.UndefinedError: 'build_number' is undefined`. More logs can be found here[1].

That is because initially we chained the `version` and `build_number` to the bucket scope, instead of the action scope. So far we could have only extracted and enriched the context with `version` and `build_number` if the release scopes were present. This failed on `jamun` as we're using `dep` bucket scopes and `push-to-candidates` action scopes. 

This PR is changing this behavior so that it's no longer chained to bucket scopes but instead to action scopes, so that any release scheduled on try, or jamun or any other low-level branch can actually be generated as long as a `build_number` and `version` are provided within the task payload. 

Currently, the only action that needs and ships a release is the `push-to-candidates`. For `push-to-nightly` we don't need them as we have `buildId` and other stuff whilst `push-to-releases` is for now done via relpro mozharness beetmover. In the near/future future we might enrich the list, but for now, it can stay as it is in my opinion.

Requesting this PR change + a beetmoverscript bump in puppet to upgrade the beetmoverworkers. This is currently affecting only jamun. It will work if we use the releasetasks graph gen for a real Fennec as the scopes there will default to 'release'.

[1]: https://public-artifacts.taskcluster.net/c5Usw0ETRhid-2WL0X9dog/0/public/logs/live_backing.log

### DIFF
--- a/beetmoverscript/utils.py
+++ b/beetmoverscript/utils.py
@@ -51,7 +51,7 @@ def write_file(path, contents):
         fh.write(contents)
 
 
-def shipping_a_release(action):
+def is_action_a_release_shipping(action):
     """Function to return boolean if we're shipping a release as opposed to a
     nightly release or something else. Does that by checking the action type.
     Currently the only action that supports release-releases is
@@ -77,7 +77,7 @@ def generate_beetmover_template_args(context):
         "platform": release_props["platform"],
     }
 
-    if shipping_a_release(context.action):
+    if is_action_a_release_shipping(context.action):
         tmpl_args["build_number"] = task['payload']['build_number']
         tmpl_args["version"] = task['payload']['version']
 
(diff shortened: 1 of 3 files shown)


beetmoverscript/utils.py
```diff
@@ -51,6 +51,16 @@ def write_file(path, contents):
         fh.write(contents)
 
 
+def shipping_a_release(action):
```
JohanLorenzo: I feel like this name doesn't return a Boolean. How about something like: `is_action_a_release_shipping`?

MihaiTabara: 👍 good suggestion, I'm following-up with a commit to change this.

PR #59, opened 2017-04-03
Bug 1344572 - add Cache-Control headers for beetmoved artifacts. r=nthomas

MihaiTabara:
…homas


nthomas-mozilla:
It looks like this would affect files beetmoved into firefox/candidates/ too. In general that's OK except we do serve RC's to the beta channel from here, so Cloudfront would hit S3 more frequently than otherwise. It's probably up to it 😄 but if we latter extend this code to other tasks a 1 hour timeout might be more of an issue.

PR #62, opened 2017-05-16
Add old id fennec platforms.

mozbhearsum:
I'm struggling to understanding how beetmover script gets its platform names when looking up manifest, but I suspect this is the cause of the beetmover signing errors on https://treeherder.mozilla.org/#/jobs?repo=date&selectedJob=98811850, which are currently using the firefox_nightly.yml manifest.

@MihaiTabara or @lundjordan, it looks like one of you are probably the best reviewers for this?


mozbhearsum:
The old-id platforms are new, and being added as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1357808. The idea is that we want to start using org.mozilla.fennec_aurora as the primary ANDROID_PACKAGE_NAME on mozilla-central, but we don't want to abandon the users that use org.mozilla.fennec. The old-id builds are a continuation of the latter, with nothing else changed. Because they are still a Fennec build I've been modeling them after "debug" platforms, hence the suffix on the platform name.

PR #65, opened 2017-05-26
Bug 1366404 - Add templates for AArch64

darchons:
Add templates for uploading the new Fennec AArch64 Nightly. The templates are copied from the corresponding x86 templates, with the architecture name replaced.

I think that's all that's needed, but I'm not sure how to actually test that.


escapewindow:
Thank you!

PR #70, opened 2017-07-06
Adjust templates for use with win32/64

srfraser:


MihaiTabara:
@srfraser want me to rollout these changes in a new beetmoverscript or do you expect to add some more stuff and then roll-out?

PR #72, opened 2017-07-15
Win templates update

srfraser:


escapewindow:
Pushed a new beetmover out. Puppet will run within the hour and it'll be live.

PR #80, opened 2017-08-31
Bug 1395649 Add support for partials

srfraser:
Provided artifacts are of the form target.partial-N.mar where currently 1 <= N <= 4
In the task's extra field there is data that allows a mapping from partial-N to the
right buildid, that lets us fill out the template.

### DIFF
--- a/beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/target.test_packages.json
+++ b/beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/target.test_packages.json
@@ -1,11 +1,11 @@
 {
     "gtest": [
-        "target.mozinfo.json",
-        "target.txt"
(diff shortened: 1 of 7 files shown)


beetmoverscript/utils.py
```diff
@@ -140,6 +142,13 @@ def get_release_props(initial_release_props_file, platform_mapping=STAGE_PLATFOR
     return update_props(props, platform_mapping)
 
 
+def get_partials_props(task):
+    '''Examine contents of task.json (stored in context.task) and extract
```
MihaiTabara: small nit: let's be consistent with using double quotes for docstrings.

beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/target.test_packages.json
```diff
@@ -1,11 +1,11 @@
 {
     "gtest": [
```
MihaiTabara: I think you can drop these changes.

PR #81, opened 2017-09-04
Version bump & commit squash

srfraser:


beetmoverscript/templates/firefox_nightly_repacks.yml
```diff
@@ -76,5 +76,12 @@ mapping:
...
         - {{ upload_date }}-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.installer.exe
         - latest-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.installer.exe
-
   {% endif %}
+  {% for partial, from_buildid in partials.iteritems() %}
+    {{ partial }}:
+      s3_key: firefox-{{ branch }}-{{ version }}-{{ platform }}-{{ locale }}-{{ from_buildid }}-{{ buildid }}.partial.mar
+      destinations:
+        - {{ upload_date }}-{{ branch }}/firefox-{{ branch }}-{{ version }}-{{ platform }}-{{ locale }}-{{ from_buildid }}-{{ buildid }}.partial.mar
+        - latest-{{ branch }}/firefox-{{ branch }}-{{ version }}-{{ platform }}-{{ locale }}-{{ from_buildid }}-{{ buildid }}.partial.mar
```
JohanLorenzo: I believe we shouldn't publish l10n mars to latest-branch. It's where en-US artifacts live.

beetmoverscript/test/__init__.py
```diff
@@ -5,12 +5,16 @@ def get_fake_valid_config():
     return load_json(path="beetmoverscript/test/fake_config.json")
 
 
-def get_fake_valid_task():
-    return load_json(path="beetmoverscript/test/test_work_dir/task.json")
+def get_fake_valid_task(taskjson='task.json'):
+    return load_json(path="beetmoverscript/test/test_work_dir/{}".format(taskjson))
+
+
+def get_fake_balrog_props_path():
```
JohanLorenzo: What's the advantage of a function instead of a constant, in this case?

srfraser: None, really, I was attempting to keep it consistent with the way the other values were created.

beetmoverscript/templates/firefox_nightly_repacks.yml
srfraser: Good catch, I will update!

PR #83, opened 2017-09-05
Bug 1396797 Support for partials in manifest

srfraser:
The outbound manifest, used by balrog scriptworker, needs to
have information about partials added.

It's unwise to add a partial if there's no related complete,
so this involves a refactor so that instead of directly
populating context.balrog_manifest, we instead fill a
staging area with details about the files we have uploaded.

Once all uploads are done, we generate the balrog manifest
to be of the expected form.

    [
        {
            "tc_nightly": true,
            "extVersion": "",
            "appVersion": "",
            "completeInfo": [
                {
                    "hash": "...",
                    "url": "...",
                    "size": 
                }
            ],
            "partialInfo": [
                {
                    "hash": "...",
                    "url": "...",
                    "size": 
                }
            ],
            "platform": "",
            "buildid": "",
            "locale": "",
            "branch": "",
            "url_replacements": [],
            "hashType": "sha512",
            "appName": "Firefox"
        }
    ]

### DIFF
--- a/beetmoverscript/script.py
+++ b/beetmoverscript/script.py
@@ -141,9 +141,9 @@ async def move_beets(context, artifacts_to_beetmove, manifest):
     # and collate it now.
     for locale in context.raw_balrog_manifest:
         balrog_entry = enrich_balrog_manifest(context, locale)
-        balrog_entry['completeInfo'] = context.balrog_manifest[locale]['completeInfo']
-        if 'partialInfo' in context.balrog_manifest[locale]:
-            balrog_entry['partialInfo'] = context.balrog_manifest[locale]['partialInfo']
+        balrog_entry['completeInfo'] = context.raw_balrog_manifest[locale]['completeInfo']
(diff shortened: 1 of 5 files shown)


beetmoverscript/script.py
```diff
@@ -120,18 +124,31 @@
...
 
             balrog_manifest = manifest['mapping'][locale][artifact].get('update_balrog_manifest')
+            # For partials
+            from_buildid = manifest['mapping'][locale][artifact].get('from_buildid')
             beets.append(
                 asyncio.ensure_future(
                     move_beet(context, source, destinations, locale=locale,
                               update_balrog_manifest=balrog_manifest,
+                              from_buildid=from_buildid,
                               artifact_pretty_name=artifact_pretty_name)
```
escapewindow: I know this predates your change, but I find it odd that we're using kwargs in the call, and args in the function. I suppose this gives us readability on the calling side and allows us to require all args on the function side?

srfraser:
I'll make a changelog when bumping the version, and merge this just now.

PR #84, opened 2017-09-06
1.0.0

srfraser:
Bump version to 1.0.0 - major version change since function call arguments have been modified. 
Add changelog


srfraser:
I see your point about the version. It was mostly after discussion with @escapewindow in https://github.com/mozilla-releng/beetmoverscript/pull/83 and irc that I made the version bump. I'll work some more on the coverage, too.

MihaiTabara:
Sounds good! 🐑 No worries about that, I meant to improve the code coverage for months but got side-tracked. I mentioned that more like something to keep ourselves in mind for the near future👍

PR #89, opened 2017-10-24
fix fennec beetmover candidates

escapewindow:
I debated lumping push-to-candidates with push-to-releases, and it looks
like the schema change busted push-to-candidates. This patch separates
them back out.

### DIFF
--- a/beetmoverscript/constants.py
+++ b/beetmoverscript/constants.py
@@ -78,7 +78,6 @@ RESTRICTED_BUCKET_PATHS = {
 # `version`
 PROMOTION_ACTIONS = (
     'push-to-candidates',
-    'push-to-releases',
 )
 RELEASE_ACTIONS = (
     'push-to-releases',
--- a/beetmoverscript/constants.py
+++ b/beetmoverscript/constants.py
@@ -76,8 +76,11 @@ RESTRICTED_BUCKET_PATHS = {
 
 # actions that imply actual releases, hence the need of `build_number` and
 # `version`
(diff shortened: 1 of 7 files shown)


beetmoverscript/constants.py
```diff
@@ -76,10 +76,13 @@
 
 # actions that imply actual releases, hence the need of `build_number` and
 # `version`
-RELEASE_ACTIONS = (
+PROMOTION_ACTIONS = (
     'push-to-candidates',
     'push-to-releases',
```
MihaiTabara: Shoudn't we remove `push-to-releases` from here and only keep it in RELEASE_ACTIONS? Otherwise, we'll have both `is_release_action` and `is_promotion_action` return True for same action scope.

escapewindow: Sure, then I would have to change https://github.com/mozilla-releng/beetmoverscript/pull/89/files#diff-14bc831c46930a709ba12c8fb4f46eb9R88 to check for both. Doable.

escapewindow:
Thanks for the quick review! This was breaking production fennec beetmover candidates. My other option was to back out the puppet version bump.

MihaiTabara:
I misread initially. Sorry about the delay with suggesting the cleanup there. We could have followed up separatelye in a different PR with the fix. Should've green flag this from start. Will pay extra care next time.

escapewindow:
No worries. I fixed it; it was simple. It probably makes more sense now :)

PR #90, opened 2017-10-26
fixes: bucket name, prefix

escapewindow:
- `bucket.objects.filter` takes a `Prefix` kwarg; I was passing in an
arg. Fixed.
- I was using `context.bucket` as the bucket name. This is `nightly` or
`dep`, which is really our bucket nickname, not the actual bucket name.
Added a `get_bucket_name` function and set `context.bucket_name` when
appropriate.

Once I installed this on beetmover-dev1 and dealt with the fact that the product in-tree is `mobile` rather than the `fennec` in `script_config.json`, we hit an [md5 mismatch](https://public-artifacts.taskcluster.net/G3SM6gCpRpuGdu--frEv3A/5/public/logs/live_backing.log), which I think is a good sign. Ideally we can either clean up pub/mobile/releases/58.0a1 , or we can bump our version in maple.


escapewindow:
@MihaiTabara , I'm currently thinking about fixing the mobile / fennec problem with a puppet patch: https://github.com/escapewindow/build-puppet/commit/a585bec548aabc3282e993da63a1e29e33509a9d

This seems a bit ugly though. I basically need pub/mobile/... and 'fennec' as a key to get the bucket name. Do you have a suggestion on how to handle this?

escapewindow:
Also, we seem to have lost our delete privs in the staging bucket - do you know anything about that?

beetmoverscript/script.py
```diff
@@ -103,6 +103,7 @@
     product = context.task['payload']['product']
     build_number = context.task['payload']['build_number']
     version = context.task['payload']['version']
+    context.bucket_name = get_bucket_name(context, product)
```
MihaiTabara: How about instead of adding it here, we add it earlier in the stream so that all the types of actions have it?  I'm thinking of something like this, [here](https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/script.py#L206).

```
#determine the S3 bucket name from config
context.bucket_name = get_bucket_name(context, context.task['payload']['product'])
```

This was we can spare both this call, which is locally to `push_to_releases` only but also the `upload_to_s3` which is used by `push_to_nightly`.

Optionally, to improve clarity, it might be worth renaming those functions too:
s/`get_task_bucket`/`get_bucket_from_scopes`
s/`get_task_action`/`get_action_from_scopes`

MihaiTabara:
Addressing the rest of the comments now, as I read them.

MihaiTabara:
I've spent some time this morning to dig a bit on this topic of `product` in the light of upcoming logic to-be-added (devedition and firefox `push-to-releases`).

I looked up to see how we solved this problem in releasetasks, where the `product` itself [comes](https://hg.mozilla.org/build/tools/file/tip/buildfarm/release/release-runner.py#l457) from Ship-it but the [root_home_dir logic](https://github.com/mozilla-releng/releasetasks/blob/master/releasetasks/__init__.py#L26) to choose between [template dirs](https://github.com/mozilla-releng/releasetasks/tree/master/releasetasks/templates) comes from [buildbot-configs](https://hg.mozilla.org/build/buildbot-configs/file/tip/mozilla/config.py#l2308), from a mapping dict. 

So far for beetmoving nightlies we didn't have to worry about this as we used the [appName ](https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/script.py#L354) from then balrog-props. But in the push-to-releases scenario we no longer have the balrog-props file so we need to rely on whatever comes from the in-tree.

Now, I think there's two ways of solving this problem:
1) Investigate why do we get `mobile` from in-tree and if there's any chance we can amend this upstream in the [worker task definition](https://hg.mozilla.org/projects/maple/file/tip/taskcluster/taskgraph/transforms/task.py#l965)?  

2) we add a mapping dict (e.g. `ALIAS_TO_PRODUCT`) either within `script_config.json`(might be easier in puppet for future development as it's easier to tweak in puppet than spin a version bump) or `constants.py` within beetmoverscript, where we define something like 

```
ALIAS_TO_PRODUCT= {
    "Fennec": "fennec",
    "mobile": "fennec",
    "devedition": "devedition",
    "firefox": "firefox",
     "Firefox": "firefox",
     .....
}
```

Having this at hand, we'd be able to use the product as in (hence `mobile` when it comes to parsing the keys `/pub/{product}/candidates` and use the `ALIAS_TO_PRODUCT[product]` when it comes to bucket names.

If we define this uniform, it might save us the string processing from [here ](https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/script.py#L352) too, and instead just call a function to return the proper value there. As soon as we start adding devedition and potentially other products, that list will grow and most likely we'll hit other mismatches with in-tree vs beetmover. (Although it seems that devedition is under the same S3 bucket as firefox so we could map it directly there)

MihaiTabara:
As to `we seem to have lost our delete privs in the staging bucket` I don't recall us ever having those credentials. Both the `bucketlister-delivery.stage` and `bucketlister-delivery.prod` buckets are managed by CloudOps so we must have lost that (if we ever had it) at some point during a cleanup/policy change.

escapewindow:
> As to we seem to have lost our delete privs in the staging bucket I don't recall us ever having those credentials. Both the bucketlister-delivery.stage and bucketlister-delivery.prod buckets are managed by CloudOps so we must have lost that (if we ever had it) at some point during a cleanup/policy change.

I don't think it makes sense to have to go to cloudops to clean up our staging envs. This will be even more clear when multiple people are pushing staging releases to try concurrently, and they all either have to create temporary directory names or versions to avoid conflicts.

If cloudops' buckets require them to clean up, have we looked into creating our own staging bucket?

beetmoverscript/script.py
escapewindow: `context.task['payload']['product']` doesn't exist outside of push-to-releases. In nightly/candidates, we get the product from the AppName in the balrog manifest. We could change that...

escapewindow:
Right now I'm thinking about specifying a fennec -> pub/mobile/ map, then specifying the product as fennec in-tree... that looks like the smallest change needed.

escapewindow:
Ok, done. I agree it might be nicest to add the product to the task definition for nightly+candidates, and then we can define `context.bucket_nick` and `context.bucket_name` in a uniform fashion early on in the script. I'm going to leave that for a future patch.

escapewindow:
https://hg.mozilla.org/projects/maple/rev/c9c7dc080e980531c1804e0f34d83069fdce8f3a for the maple mobile->fennec

MihaiTabara:
> I don't think it makes sense to have to go to cloudops to clean up our staging envs. This will be even more clear when multiple people are pushing staging releases to try concurrently, and they all either have to create temporary directory names or versions to avoid conflicts. If cloudops' buckets require them to clean up, have we looked into creating our own staging bucket?

Yes and yes. I agree we should have access to doing that or a separate bucket to use.

> Ok, done. I agree it might be nicest to add the product to the task definition for nightly+candidates, and then we can define context.bucket_nick and context.bucket_name in a uniform fashion early on in the script. I'm going to leave that for a future patch.

👍 

Also good catch on this fix https://hg.mozilla.org/projects/maple/rev/c9c7dc080e980531c1804e0f34d83069fdce8f3a

beetmoverscript/script.py
MihaiTabara: Filed https://github.com/mozilla-releng/beetmoverscript/issues/91 to track this.

PR #93, opened 2017-11-03
Bug 1405622 - Stop publishing PMARs under latest dirs and assign dedicated subdir.

MihaiTabara:
We've reached a resolution under [bug 1405622](https://bugzilla.mozilla.org/show_bug.cgi?id=1405622) to:

- stop publishing partial mars under latest directories for both `en-US` and `locales`
- have a dedicated subdir for these instead to ease the implementation of a lifecycle policy 
- locale partial.mar are going in the `date-branch-l10n` subdirs as opposed to `date-branch` where `en-US` lives

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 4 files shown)


MihaiTabara:
Approved by @nthomas-mozilla in bugzilla as well.

PR #94, opened 2017-11-15
Bug 1417568 - beetmoverscript push-to-cdns should not silently return success if nothing has been copied over

MihaiTabara:


MihaiTabara:
@escapewindow 

I am thinking of merging this to master branch but not bumping beetmoverscript. It's not urgent and can be picked up by the next set of changes. I'm just trying to minimize the amount of change freeze emails I'm sending :) 

Therefore, please ignore the version.txt changes.

PR #96, opened 2017-12-09
Bug 1423885 - Yet another milestone beetmoverscript 

MihaiTabara:
* fixes checksums
* fixes devedition
* misc


CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.5.0] - 2017-12-12
+
+### Added
+- beetmoverscript support for Devedition releases
```
escapewindow: We seem to also have added `.beet` support in general, no?
Also, `get_product_name`, `NORMALIZED_BALROG_PLATFORMS`.

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
+
+### Changed
+- stop uploading checksums.asc files as `.beet`  under beetmover-checksums
+- `get_release_props` and `update_props`  functions now take context as argument
```
escapewindow: This is a breaking change. It's not as critical for beetmoverscript to respect semver since we have nothing downstream. But if we follow semver, we would bump the major version. 4.0.0.

beetmoverscript/utils.py
```diff
@@ -134,18 +147,26 @@ def update_props(props, platform_mapping):
     # l10n jobs have it set under `platform`. This is merely an uniformization
     # under the `stage_platform` field that is needed later on in the templates
     stage_platform = props.get("stage_platform", props.get("platform"))
+    # XXX Bug 1424482 - until we solve this, we need this hack. Since en-US
+    # have at least `stage_platform`, there is a way to tell whether they are
+    # devedition related or not. But for l10n jobs, we only have `platform`
+    # which is identical to the ones we have for Firefox.
```
escapewindow: As mentioned in person, we can use graph generation to take `attributes['shipping_product']` and turn it into a `task.payload.product`; similarly we could use `build_platform` and turn it into a `task.payload.platform`. If we accept those as optional task payload schema items, and keyed off of them if they exist, then we could slowly reduce the need for `balrog_props.json`.

PR #98, opened 2017-12-14
Clearer logs of what's moving where

srfraser:
Should we add more, or is this enough to get on with?


MihaiTabara:
This looks excellent. Let's start with this. I'll merge and bump along with my changes to fix checksums. If we need more, we can always roll-out another beetmoverscript version.

Thank you for your help!

PR #101, opened 2018-02-15
Bug 1424482 - Support task.payload.releaseProperties

JohanLorenzo:


beetmoverscript/task.py
```diff
@@ -4,11 +4,15 @@
 import re
 import shutil
 import scriptworker.client
+
+from copy import deepcopy
+
+from beetmoverscript import utils
```
MihaiTabara: Am curious, do you have a preference of using `submodule.function` for the ease of reading or is there another reason?

beetmoverscript/task.py
```diff
@@ -134,6 +138,27 @@ def get_upstream_artifacts(context):
...
+    initial_release_props_file = None
+
+    if payload_properties:
+        props = payload_properties
+        log.debug("Loading release_props from task's payload: {}".format(props))
+    else:
+        initial_release_props_file = get_initial_release_props_file(context)
+        props = utils.load_json(initial_release_props_file)['properties']
+        log.warn(
+            'Deprecated behavior! This will be gone after Firefox 59 reaches release. Loading release_props from "{}": {}'
```
MihaiTabara: Tiny nit: maybe split here the `Loading ...` part on a newline to cope with PEP8.

beetmoverscript/task.py
```diff
@@ -134,6 +138,27 @@ def get_upstream_artifacts(context):
...
+    payload_properties = context.task.get('payload', {}).get('release_properties', None)
+    initial_release_props_file = None
+
+    if payload_properties:
+        props = payload_properties
+        log.debug("Loading release_props from task's payload: {}".format(props))
+    else:
+        initial_release_props_file = get_initial_release_props_file(context)
+        props = utils.load_json(initial_release_props_file)['properties']
+        log.warn(
```
MihaiTabara: ++ for the `log.warn()`

beetmoverscript/task.py
```diff
@@ -134,6 +138,27 @@ def get_upstream_artifacts(context):
...
+            'Deprecated behavior! This will be gone after Firefox 59 reaches release. Loading release_props from "{}": {}'
+            .format(initial_release_props_file, props)
+        )
+
+    final_props = update_props(context, props, platform_mapping)
```
MihaiTabara: This call should be called only in the `else` statement above. 
The `update_props()` was a hack we added at some point to map correctly some fields like: `platform`, `stage_platform` and alike. To be honest, I'm hoping all this goes away in the new implementation as we can nicely and easily set those up in-tree. Now that we have `shipping_phase` and `shipping_product` we can infer all these previously-hacked values via transforms in-tree. So we don't need to perform all these hacks within beetmoverscript.

For context, the reason why all these hacks are here, is because the `balrog_props` file, generated at build time, doesn't look the same across products and platforms. It's missing `stage_platform` in some products, but is present in others. Likewise for `platform` or `locale`. But in-tree, we can finally standardize and append all these fields so that we can finally clean this up.

beetmoverscript/task.py
JohanLorenzo: Actually, that's because of the monkeypatch used in the unittests. I wish we could use `from module.submobile import function`. If we define so, then momnkeypatch fails to patch. For more info: https://github.com/pytest-dev/pytest/issues/2229

beetmoverscript/task.py
JohanLorenzo: Why not. FWIW, we had a discussion about line length a bit less than a year ago. The max line length used within releng projects is 160. For reference: https://github.com/mozilla-releng/services/pull/282

beetmoverscript/task.py
JohanLorenzo: That's good to know. Do you want me to make the changes in this PR or a follow would be good enough?

beetmoverscript/task.py
MihaiTabara: Ah, okay, didn't know this.  Thanks for pointing this out!

beetmoverscript/task.py
MihaiTabara: 👍

beetmoverscript/task.py
MihaiTabara: Either wfm, we can address it later on too, just not to forget of it.

JohanLorenzo: Okay, I'll go with a follow up. I'm not familiar enough with beetmover to be sure it's okay to remove these pieces of data

JohanLorenzo: Follow up at #104

PR #104, opened 2018-02-28
Use update_props only when data comes from balrog_props.json

JohanLorenzo:
Follow up #101.

@MihaiTabara, if we remove this function call, platform (not the staging one) becomes wrong. See tests for more details. What do you think I should do in this case?


MihaiTabara:
I think you're right, we don't need this. I looked again last night and we might not need that call as I said in the other PR. That's because 59.0 will still rely on balrog props whereas 60+ will be payload-relprops based. Both need the `update_props` so I think it's safe to leave it as is.

JohanLorenzo:
Sounds good to me. Thank you for the explanation!

PR #105, opened 2018-03-06
Bug 1443588: Refactor the fennect templates, so that they don't vary by platform

tomprince:
Before, the only essential difference between them is the `android-*`
slug in the file names. Make this a template parameter, so they can be
merged.


beetmoverscript/task.py
```diff
@@ -62,7 +62,7 @@ def get_task_action(task, script_config):
         messages.append("Only one action type can be used")
 
     action = actions[0]
-    if action not in script_config['actions']:
+    if action not in script.action_map:
```
escapewindow: As a constant, `action_map` should probably be in all caps... not your bug but might be a good fix.

PR #106, opened 2018-03-14
bug 1432591 - Source tarball

escapewindow:
We're going back to the source tarball approach.
I managed to get this working on maple, except for the missing beetmover manifest patch.


escapewindow:
Thanks! I'll roll this out to the dev pool first.

PR #107, opened 2018-03-15
bug 1432591 - use pretty naming.

escapewindow:
There *might* be another followup patch to add the bundle, depending on the responses to https://bugzilla.mozilla.org/show_bug.cgi?id=1432591#c40 .

### DIFF
--- a/beetmoverscript/templates/fennec_candidates.yml
+++ b/beetmoverscript/templates/fennec_candidates.yml
@@ -127,13 +127,13 @@ mapping:
       destinations:
         - SOURCE.asc
     source.tar.xz:
-      s3_key: source.tar.xz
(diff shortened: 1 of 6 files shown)


beetmoverscript/templates/fennec_candidates.yml
```diff
@@ -127,13 +127,13 @@ mapping:
       destinations:
         - SOURCE.asc
     source.tar.xz:
-      s3_key: source.tar.xz
+      s3_key: source/fennec-{{ version}}.source.tar.xz
```
MihaiTabara: I don't think we even offered this before for Fennec, so it's a nice win along the way (e.g. https://archive.mozilla.org/pub/mobile/candidates/58.0b3-candidates/build4/)

PR #111, opened 2018-03-21
Bug 1403548: Support linux64-asan-reporter platform

catlee:


MihaiTabara:
If you need to roll this out, we also need to bump version. Might want to talk to @JohanLorenzo, he was about to rollout something already in https://github.com/mozilla-releng/beetmoverscript/pull/110 for some TB testing. Otherwise, I can take care of this too.

PR #112, opened 2018-03-21
Bug 1447673 - first draft to fix EME-free partner repacks.

MihaiTabara:


MihaiTabara:
Assuming the following three things happen in-tree for partner repacks (at least eme-free ones so far)
1. switch action from `push-to-candidates` to `push-to-partner`
2. fix the locale in upstreamArtifacts which now says `locale = eme-free` instead of `en-US` or w\e other locale
3. I assume we're attempting to make `en-US` work for now but after that we're unthrottling the rest of locales too.

config_example.json
```diff
@@ -28,7 +28,10 @@
             "devedition_candidates_repacks": "/path/to/beetmoverscript/beetmoverscript/templates/devedition_candidates_repacks.yml"
         },
         "push-to-releases": {},
-        "push-to-staging": {}
+        "push-to-partner": {
+            "firefox_partner": "/path/to/beetmoverscript/beetmoverscript/templates/firefox_partner.yml",
+            "firefox_partner_repacks": "/path/to/beetmoverscript/beetmoverscript/templates/firefox_partner_repacks.yml"
```
mozbhearsum: I don't know this code, but I suspect that we don't need both of these. Partner repacks don't happen until after en-US + l10n is done, and all locales (including en-US) are available at the same time.

escapewindow: Hm, possible, especially if we create the payload so `en-US` is an explicit locale, rather than assumed.

beetmoverscript/templates/firefox_partner.yml
```diff
@@ -0,0 +1,42 @@
+---
+metadata:
+    name: "Beet Mover Manifest"
+    description: "Maps Firefox artifacts to pretty names for en-US"
+    owner: "release@mozilla.com"
+
+s3_bucket_path: pub/firefox/candidates/{{ version }}-candidates/build{{ build_number }}/
```
escapewindow: Hm, is the private partners path different? If so, it would be nice to be able to reuse the template, but change the `s3_bucket_path` if the private flag is set.

beetmoverscript/templates/firefox_partner.yml
```diff
@@ -0,0 +1,42 @@
...
+
+mapping:
+{% for locale in ['en-US'] %}
+  {{ locale }}:
+  {% if platform in ["linux-i686", "linux-x86_64"] %}
+    target.tar.bz2:
+      s3_key: {{ platform }}-EME-free/{{ locale }}/firefox-{{ version }}.tar.bz2
```
escapewindow: Ah, looks like this is currently the public template only. This works, but we'll need to either support both the private and public partners in the same template(s), or have separate partners behaviors and templates.

config_example.json
```diff
@@ -28,7 +28,10 @@
             "devedition_candidates_repacks": "/path/to/beetmoverscript/beetmoverscript/templates/devedition_candidates_repacks.yml"
         },
         "push-to-releases": {},
-        "push-to-staging": {}
+        "push-to-partner": {
```
escapewindow: As stated above, we'll either need to support another behavior (`push-to-partner-public` vs `push-to-partner-private`?) or we'll need templates that can handle the various partner files.

escapewindow:
I'm leaning towards keeping this a dev branch until we're ready to roll out, but I'm also open to merging and itererating on master (we can branch off an earlier release if we need to make a prod dot release sooner).

MihaiTabara:
I have more stuff locally + WIP on tests to get back to 100% but for now, I wanted to make sure I'm on the right track.

beetmoverscript/script.py
```diff
@@ -283,6 +301,70 @@ def list_bucket_objects(context, s3_resource, prefix):
...
+    if is_partner_private_task(context):
+        elements = pretty_full_path.split('/')
+        identifier = '{version}-{build_number}'.format(version=version, build_number=build_number)
+        # we need need to manually insert the "version-buildno" identifier in
+        # between `partner` # and `partner-variant` to be consistent
+        elements.insert(1, identifier)
+        # TODO: potentially need to remove the `v1` from the path?
+        path = '/'.join(elements)
+        # XXX: temp hack until bug 1447673 is solved
+        if context.bucket == "dep":
```
escapewindow: hm, won't the private and public uploads upload to the same place? Doesn't matter?

MihaiTabara: Sadly it does. Absent this hack (which we need while we test on maple/staging env), the
* private tasks will try to upload to http://ftp.stage.mozaws.net/$partner/$partner-variant ... directly in the root of the bucket which I'm not sure a) we're allowed b) we should 
* public tasks will go under regular candidates/partner-repacks/$partner/$partner-variant

PR #113, opened 2018-03-27
Bug 1403548: Add asan platforms to platform specific nightly templates

catlee:


MihaiTabara:
Maybe worth adding a note of this asan reporter thing in `Changelog` since we're gonna unroll another version.

PR #119, opened 2018-04-12
use `locale` as the path in partner repacks.

escapewindow:
This should hopefully get our beetmover partner repacks working.

Originally, we thought we would write the artifacts to a path, and preserve that relative path when beetmoving. However, public and private paths differ.

Instead, I'm overloading the `locale` attribute for the path to use. We should probably do this in a different way, but it looks like it's working so far.

Creating the PR for coverage testing, since the `get_hash` test seems to break on me locally on osx.


escapewindow:
- i want to add more coverage
- I'm a little worried about not sanity checking the paths at all, and only relying on the tree to do the right thing.

beetmoverscript/constants.py
```diff
@@ -155,6 +158,11 @@
...
 
-PARTNER_LEADING_STRING = 'releng/partner/'
-PARTNER_REPACK_PUBLIC_PREFIX_TMPL = 'pub/firefox/candidates/{version}-candidates/build{build_number}/partner-repacks/'
-PARTNER_REPACK_PUBLIC_PAYLOAD_ID = 'is_partner_repack_public'
+PARTNER_REPACK_PUBLIC_PREFIX_TMPL = 'pub/firefox/candidates/{version}-candidates/build{build_number}/'
+PARTNER_REPACK_PRIVATE_REGEXES = (
+    r'^[^\/.]+/{version}-{build_number}/[^\/.]+\/(mac|win32|win64|linux-i686|linux-x86_64)/[^\/.]+$',
+)
+PARTNER_REPACK_PUBLIC_REGEXES = (
+    r'^(mac|win32|win64)-EME-free\/[^\/.]+$',
```
MihaiTabara: Ah, had this on my TODO list and forgot to mention it. I assumed we could trim that `v1` from the first iteration. This is much better though 👍

beetmoverscript/script.py
```diff
@@ -335,30 +346,24 @@ def get_destination_for_private_repack_path(context, manifest, full_path, locale
 
     # pretty name the `target` part to the actual filename
     pretty_full_path = os.path.join(
-        os.path.dirname(full_path),
+        locale,
```
MihaiTabara: Yeah, I agree, stashing this within `locale` seems like a hack for whoever reads the code for the first time but I think it's good enough for now to solve the problem. I'm curious how this is handled in the generate_beetmover-manifest section where the `locale` is essential but I'll glance in a bit at that particular part.

beetmoverscript/script.py
```diff
@@ -335,30 +346,24 @@ def get_destination_for_private_repack_path(context, manifest, full_path, locale
...
     if is_partner_private_task(context):
-        elements = pretty_full_path.split('/')
-        identifier = '{version}-{build_number}'.format(version=version, build_number=build_number)
-        # we need need to manually insert the "version-buildno" identifier in
-        # between `partner` # and `partner-variant` to be consistent
-        elements.insert(1, identifier)
-        # TODO: potentially need to remove the `v1` from the path?
-        path = '/'.join(elements)
-        # XXX: temp hack until bug 1447673 is solved
-        if context.bucket == "dep":
```
MihaiTabara: So happy this goes away.

beetmoverscript/test/test_script.py
```diff
@@ -509,22 +511,54 @@ def test_get_destination_for_private_repack_path(context, full_path, payload_id,
...
-            get_destination_for_private_repack_path(context, mapping_manifest,
+            get_destination_for_partner_repack_path(context, mapping_manifest,
                                                     full_path, locale)
     else:
-        assert expected == get_destination_for_private_repack_path(context, mapping_manifest,
+        assert expected == get_destination_for_partner_repack_path(context, mapping_manifest,
                                                                    full_path, locale)
 
 
+# sanity_check_partner_path {{{1
```
MihaiTabara: I like all these tests! :)

beetmoverscript/test/test_utils.py
```diff
@@ -15,14 +15,14 @@
...
-from beetmoverscript.constants import HASH_BLOCK_SIZE, PARTNER_REPACK_PUBLIC_PAYLOAD_ID
+from beetmoverscript.constants import HASH_BLOCK_SIZE
 
 assert context  # silence pyflakes
 
 
 # get_hash {{{1
 def test_get_hash():
-    correct_sha1 = 'cb8aa4802996ac8de0436160e7bc0c79b600c222'
+    correct_sha1s = ('cb8aa4802996ac8de0436160e7bc0c79b600c222', 'da39a3ee5e6b4b0d3255bfef95601890afd80709')
```
MihaiTabara: Heh, nice one :)

beetmoverscript/utils.py
```diff
@@ -86,18 +85,14 @@ def is_partner_private_task(context):
...
 
 
 def is_partner_public_task(context):
     """Function to return boolean if we're considering a private partner task.
     Does that by checking the action type and absence of a flag in payload
     """
-    task = context.task
-    return (is_partner_action(context.action) and
-            PARTNER_REPACK_PUBLIC_PAYLOAD_ID in task['payload'])
+    return (is_partner_action(context.action) and 'partner' not in context.bucket)
```
MihaiTabara: Ah, smart. Not sure why I haven't thought of that. Makes more sense and is much simplified!

beetmoverscript/constants.py
```diff
@@ -155,6 +158,11 @@
     'thunderbird': 'pub/thunderbird/',
 }
 
-PARTNER_LEADING_STRING = 'releng/partner/'
-PARTNER_REPACK_PUBLIC_PREFIX_TMPL = 'pub/firefox/candidates/{version}-candidates/build{build_number}/partner-repacks/'
-PARTNER_REPACK_PUBLIC_PAYLOAD_ID = 'is_partner_repack_public'
+PARTNER_REPACK_PUBLIC_PREFIX_TMPL = 'pub/firefox/candidates/{version}-candidates/build{build_number}/'
+PARTNER_REPACK_PRIVATE_REGEXES = (
+    r'^[^\/.]+/{version}-{build_number}/[^\/.]+\/(mac|win32|win64|linux-i686|linux-x86_64)/[^\/.]+$',
```
nthomas-mozilla: The `.../[^\/.]+\/...` part in the middle of PARTNER_REPACK_PUBLIC_REGEXES doesn't like partner subconfigs like mail.com_notb in unitedinternet ([eg](https://tools.taskcluster.net/groups/I_lv8okdSXqL-1J6o_xgGQ/tasks/dhSha55MQhOGwFet-f8dkA/details)). Maybe replace it with a `.../[^\/]+?/...`  (ie non-greedy matching allowing periods, and remove what looks like a stray escape the following /).

escapewindow: I added `.` to disallow `/../` patterns. I can try to change the regex and checks to allow for a normal usage of periods, but not `..`

PR #122, opened 2018-04-17
Use the URL appropriate for the bucket in balrog manifests.

tomprince:


config_example.json
```diff
@@ -51,7 +54,8 @@
             },
             "buckets": {
                 "firefox": "mozilla-releng-private-firefox-bucket"
-            }
+            },
+            "url_prefix": "https://archive.mozilla.org"
```
tomprince: I'm not sure what this should be.

tomprince:
I think this is ready for review.

beetmoverscript/utils.py
```diff
@@ -253,3 +253,7 @@ def get_creds(context):
 
 def get_bucket_name(context, product):
     return context.config['bucket_config'][context.bucket]['buckets'][product]
+
+
+def get_bucket_url_prefix(context):
+    return context.config['bucket_config'][context.bucket]['url_prefix']
```
escapewindow: Hm, I suppose it's a feature that this will die if we try to push nightlies to the partner bucket. I'm wondering if we should add a schema for the `script_config.json` at some point.

PR #124, opened 2018-04-22
  Add automatic changelog generation support.

tomprince:
Currently, [towncrier](https://github.com/hawkowl/towncrier/) only supports reStructued text, so this switches the changelog to that format.


beetmoverscript/newsfragments/124.added
```diff
@@ -0,0 +1 @@
+Add automatic changelog generation using [towncrier](https://github.com/hawkowl/towncrier/).
```
escapewindow: Do we add these newsfragments manually? What do we name them if there's no related issue?

pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[[tool.towncrier.type]]
+directory = "removed"
+name = "Removed"
+showcontent = true
+
+[[tool.towncrier.type]]
+directory = "changed"
+name = "Changed"
+showcontent = true
```
escapewindow: Should we add a `fixed` ?

tomprince: https://github.com/mozilla-releng/beetmoverscript/pull/126

PR #129, opened 2018-04-26
Bug 1456234: Fix support for fennec  sources to be on `*-release` platform.

tomprince:


tomprince:
We probably only need in case we need to do a 59 chem spill. But it doesn't hurt us to leave compatibility in for a cycle. The change to use source packages again was uplifted in an early 60 beta.

PR #130, opened 2018-04-30
 Bug 1446815 - add custom flexibility for checksums filename 

MihaiTabara:
Work in progress to add source files within the checksums for each release.


beetmoverscript/task.py
```diff
@@ -87,9 +88,16 @@ def generate_checksums_manifest(context):
     return '\n'.join(content)
 
 
+def is_custom_beetmover_task(context):
```
escapewindow: Do we plan on using this function outside of checksums? If not, adding `checksums` somewhere in the name may help clarify its purpose.

MihaiTabara: Good point, I've addressed it now in a follow-up commit: https://github.com/mozilla-releng/beetmoverscript/pull/130/commits/c7d542e8ae5d44db0597b0b8a603c01059112262

PR #132, opened 2018-05-03
Fix fennec source platform for 60.

tomprince:
Apprarently, `android-api-16` is a possible platform there.


MihaiTabara:
I'll merge this and test locally and then push it out of the door

PR #140, opened 2018-05-17
Drop py35 support; add py37 tests

lisajguo:


escapewindow:
We may want to update https://github.com/mozilla-releng/beetmoverscript/blob/master/setup.py#L44 to say 3.6, otherwise lgtm!

PR #146, opened 2018-05-24
7.2.3

tomprince:


MihaiTabara:
Happy to see `newfragments` files flushing away, AFAIUI, they're not supposed to be upstream - https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/newsfragments/143.fixed 👍

tomprince:
They are supposed to collect between releases (see [here](https://github.com/twisted/twisted/tree/trunk/src/twisted/newsfragments) for example). Usually, every PR should have a newsfragement.

PR #149, opened 2018-05-29
stripped balrog props

lisajguo:


lisajguo:
It's all passing now except for py37, which catlee said is okay since we're not rolling out 37 any time soon

beetmoverscript/utils.py
```diff
@@ -119,7 +119,6 @@ def generate_beetmover_template_args(context):
         "branch": release_props["branch"],
         "product": release_props["appName"],
         "stage_platform": release_props["stage_platform"],
-        "platform": release_props["platform"],
```
MihaiTabara: Hm, I don't think this will work. Removing `platform` will cause templates to fail. There's plenty of references to "{{ platform }} in all over the templates from https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/templates/ (for example https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/templates/firefox_candidates.yml#L13 )

This is another reason why I should've added integration tests to this repo long time ago, bah.
@lisajguo can you please try to run beetmover locally and confirm this works or not? Something similar to what we've done together on Monday. Please test with a `beetmover-repackage` and `beetmover-checksums` beetmover jobs.

beetmoverscript/task.py
```diff
@@ -148,31 +132,12 @@ def get_release_props(context, platform_mapping=STAGE_PLATFORM_MAP):
     """determined via parsing the Nightly build job's balrog_props.json and
```
MihaiTabara: We need to fix the docstring too, since it no longer reflects `balrog_props`. See my below comment

beetmoverscript/task.py
```diff
@@ -148,31 +132,12 @@ def get_release_props(context, platform_mapping=STAGE_PLATFORM_MAP):
...
-        log.warn(
-            'Deprecated behavior! This will be gone after Firefox 59 reaches release. Loading release_props from "{}": {}'
-            .format(initial_release_props_file, props)
-        )
-
-    final_props = update_props(context, props, platform_mapping)
-    return (final_props, initial_release_props_file)
 
+    # if payload_properties:
+    props = payload_properties
```
MihaiTabara: This looks good; however, I think we can make two small improvements:

1. Since we're dropping the balrog_props behavior, it means the release_properties field in `task.payload` is mandatory. So I think it's worth barfing if those values are not set. 
2. I don't really see the value of `props` here, since it's just another reference to `payload_properties`.

That being said, I'd rewrite the function to be something like:
```
def get_release_props(context, platform_mapping=STAGE_PLATFORM_MAP):
    """Function to grab the release properties data from task payload"""
    payload_properties = context.task.get('payload', {}).get('releaseProperties', None)
    if not payload_properties:
        raise ScriptWorkerTaskException(
            "could not determine release props file from task payload"
        )

    log.debug("Loading release_props from task's payload: {}".format(payload_properties))
    return update_props(context, payload_properties, platform_mapping)

```

beetmoverscript/utils.py
lisajguo: Yep, on it

beetmoverscript/task.py
lisajguo: good catch!

lisajguo:
I used mach to generate the taskgraph as you said and balrog_props.json is nowhere to be seen so It think we should be okay to merge this PR.

beetmoverscript/data/beetmover_task_schema.json
```diff
@@ -49,6 +49,9 @@
                         "hashType":  {
                             "type" : "string"
                         },
+                        "stage_platform":  {
+                            "type" : "string"
+                        },
```
MihaiTabara: Hm, I don't get this, why is this needed? And if so, is it required or optional?

beetmoverscript/task.py
MihaiTabara: We still need to amend the docstring here

beetmoverscript/data/beetmover_task_schema.json
lisajguo: By my understanding, stage_platform is a hack that maps stage platform names and is needed in releaseProperties for the time being https://github.com/lisajguo/beetmoverscript/blob/master/beetmoverscript/constants.py#L15

It's optional because: https://github.com/lisajguo/beetmoverscript/blob/master/beetmoverscript/task.py#L165

lisajguo: Will update the schema to explicitly state that it's optional

MihaiTabara:
I branched-off this PR in another PR in order to attempt to fix it. I'll close this for now in favor of https://github.com/mozilla-releng/beetmoverscript/pull/153

PR #150, opened 2018-06-01
Create checksums for public partner builds

nthomas-mozilla:


beetmoverscript/script.py
```diff
@@ -318,6 +321,12 @@ def list_bucket_objects(context, s3_resource, prefix):
                     upload_to_s3(context=context, s3_key=destination, path=source)
                 )
             )
+
+            if is_partner_public_task(context):
+                context.checksums[full_path_artifact] = {
+                    algo: get_hash(source, algo) for algo in context.config['checksums_digests']
+                }
+
```
MihaiTabara: I think you also need something like `context.checksums[artifact_pretty_name]['size'] = get_size(source)` too like [here](https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/script.py#L293) as it's being used later on when checksums file is being generated [here](https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/task.py#L84)

nthomas-mozilla: Thanks for catching that.

nthomas-mozilla:
This seems to be working fine, eg https://tools.taskcluster.net/groups/Z9TwFwytTPq1ZngU-8IqGw/tasks/CD0J_zlGSHaICX5upL2vTg/runs/0/artifacts. I'll wait and see if there are any more tweaks I need [bug 1422471](https://bugzilla.mozilla.org/show_bug.cgi?id=1422471) before requesting review.

nthomas-mozilla:
This is working in a sort-of staging release, more details [in bug 1422471 comment 5](https://bugzilla.mozilla.org/show_bug.cgi?id=1422471#c5), and is ready for review.

I've not implemented signing of target.checksums on account of not using them when generating checksums, and it's another 285 tasks to add to the graph.

Since you last saw this I realized the filenames were wrong in the beet files, so I'm munging the s3 destination (recall locale is set to something like `beetmover-checksums/mac-EME-free/de` in the upstreamTasks definition). Also needed to allow public partner files to go into beetmover-checksums.

beetmoverscript/script.py
```diff
@@ -313,6 +316,16 @@ def list_bucket_objects(context, s3_resource, prefix):
                     upload_to_s3(context=context, s3_key=destination, path=source)
                 )
             )
+
+            if is_partner_public_task(context):
+                # we trim the full destination to the part after
+                # candidates/61.0b2-candidates/buildN/
+                artifact_pretty_name = destination[destination.find(locale):]
```
MihaiTabara: Nit, not blocking: I wonder if it makes sense to guard this block code with something like https://github.com/mozilla-releng/beetmoverscript/pull/150/files#diff-98ba072d44cc67fadf38dc459a83d934R287 in order to make sure we're not pushing a duplicate in that dict. Most likely the chances are minimal for that to happen but it doesn't harm to add it maybe.

beetmoverscript/script.py
```diff
@@ -313,6 +316,16 @@ def list_bucket_objects(context, s3_resource, prefix):
                     upload_to_s3(context=context, s3_key=destination, path=source)
                 )
             )
+
+            if is_partner_public_task(context):
+                # we trim the full destination to the part after
+                # candidates/61.0b2-candidates/buildN/
```
MihaiTabara: Nit: let's make this comment generic like `candidates/{version}-candidates/buildN/` for the ease of reading.

PR #155, opened 2018-06-06
Buildhub

lisajguo:
Unfinished patch. Pushing to get feedback if I'm on the right track, but I have an unanswered question about where to grab data from to update the fields https://bugzilla.mozilla.org/show_bug.cgi?id=1443873#c8

Am I following the correct workflow? This seems unwieldy, what I'm doing.


lisajguo:
I wrote this patch with the assumption that buildhub.json will be received as an upstream artifact by beetmoverscript

MihaiTabara:
Removing @JohanLorenzo from reviewers until we're in a more stable state with the PR

beetmoverscript/script.py
```diff
@@ -248,6 +248,13 @@ def list_bucket_objects(context, s3_resource, prefix):
         for artifact in artifacts_to_beetmove[locale]:
             source = artifacts_to_beetmove[locale][artifact]
             artifact_pretty_name = manifest['mapping'][locale][artifact]['s3_key']
+
+            # update buildhub.json fields
+            if artifact == 'buildhub.json':
+                write_json(source, update_buildhub_json(source, context, manifest))
+                # no pretty name for buildhub.json
```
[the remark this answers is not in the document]
lisajguo: ah I see! makes sense.

MihaiTabara:
We now have coveralls enabled again in this repo. Let's test this in staging as soon as we're done with the in-tree changes.

MihaiTabara:
I triggered https://tools.taskcluster.net/groups/CujlHlj8Rg2YhWkfIPu7YQ to test the recent patches.

MihaiTabara:
Rolled-out another beetmoverscript version out of the tip of this. Rerunning the beetmover-reackage-{platform} jobs now.

MihaiTabara:
Rolled-out 7.5.0.dev3 to test the tip of this PR.

MihaiTabara:
https://tools.taskcluster.net/groups/P45R5UvjS_-GX66dsMyw_g is tracking tip of this PR

MihaiTabara:
I rolled-out another version for tip of this PR and rerun all the `beetmover-repackage` tasks.

beetmoverscript/script.py
```diff
@@ -247,10 +249,32 @@ def list_bucket_objects(context, s3_resource, prefix):
...
 async def move_beets(context, artifacts_to_beetmove, manifest):
     beets = []
+
     for locale in artifacts_to_beetmove:
+
+        installer_path = ''
+
+        # get path of installer beet
+        for artifact in artifacts_to_beetmove[locale]:
+            # import pdb; pdb.set_trace()
```
MihaiTabara: Remove this line. Debugging lines, commented or not are not supposed to get into production.

beetmoverscript/script.py
```diff
@@ -247,10 +249,32 @@ def list_bucket_objects(context, s3_resource, prefix):
...
+            if artifact in INSTALLER_ARTIFACTS:
+                installer_path = artifacts_to_beetmove[locale][artifact]
+
+        # error check
+        if not installer_path:
```
MihaiTabara: Hm, this is wrong. This `move_beets` function is (re)used by other beetmover jobs (e.g. `beetmover-checksums`, `beetmover-source`, `beetmover-signed-langpacks-nightly`, etc), all of which don't have the installer within their `upstreamArtifacts`. They will barf at this check and fail. I glanced at the last staging releases and I can confirm that happens. See https://taskcluster-artifacts.net/LXvVyIerRmCEfuCWbrUUQw/0/public/logs/live_backing.log (one of the errors logs for `beetmover-checksums). Or just glance at the staging releases "Failed" tasks and there are many.

What you can do here is a) check if buildhub artifact is present in the upstreamArtifacts. b) check if the installer is present. You can then raise exception if buildhub is present but not the installer path.

beetmoverscript/task.py
```diff
@@ -1,14 +1,19 @@
...
+import urllib.parse
 
 from copy import deepcopy
 from scriptworker import client
 
 from beetmoverscript import utils, script
-from beetmoverscript.constants import (STAGE_PLATFORM_MAP,
-                                       RESTRICTED_BUCKET_PATHS,
-                                       CHECKSUMS_CUSTOM_FILE_NAMING)
+
```
MihaiTabara: This blank line is not needed. These are local application/library specific imports and according with the [PEP8 documentation](https://www.python.org/dev/peps/pep-0008/#imports), should be chunked together.

beetmoverscript/task.py
```diff
@@ -152,3 +157,25 @@ def update_props(context, props, platform_mapping):
...
+    url_prefix = context.config["bucket_config"][context.bucket]["url_prefix"]
+
+    # Update fields
+    contents['download']['size'] = utils.get_size(installer_path)
+    contents['download']['date'] = str(arrow.utcnow())
+    
+    url = os.path.join(manifest["s3_bucket_path"], dest[0])
+    # URL encode the url; we are assuming the url prefix is ascii safe
+    url = '/'.join([urllib.parse.quote(x) for x in url.split('/')])
+    contents['download']['url'] = url_prefix + "/" + url
```
MihaiTabara: I'd rather join these two strings with the help of a function, rather than this. I'm always feeling unsafe when doing these operations by hand using string concatenation. We can maybe use [urllib.parse.urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) or simply the `os.path.join` from line 176 or something.

beetmoverscript/templates/devedition_candidates_repacks.yml
```diff
@@ -19,6 +19,10 @@ mapping:
       s3_key: {{ platform }}/xpi/{{ locale }}.xpi
       destinations:
         - {{ platform }}/xpi/{{ locale }}.xpi
+    buildhub.json:
```
MihaiTabara: We don't yet have the support for l10n in generating the buildhub.json so this will be useless for now. But I think we'll end up adding it later anyway once bug 1459302 is solved. So I think we can leave these 👍

beetmoverscript/templates/fennec_nightly.yml
```diff
@@ -128,5 +128,9 @@ mapping:
       destinations:
         - latest-{{ branch }}-{{ platform }}/{{ locale_prefix }}robocop.apk
         - {{ upload_date }}-{{ branch }}-{{ platform }}/{{ locale_prefix }}robocop.apk
-
+    buildhub.json:
+      s3_key: latest-{{ branch }}-{{ platform }}/{{ locale_prefix }}fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.buildhub.json
```
MihaiTabara: `s3_key` is solely used in checksums generation. Whatever we end up placing here, will be landing in the SHA512SUMS files (e.g. [here](https://archive.mozilla.org/pub/firefox/candidates/61.0b14-candidates/build1/SHA256SUMS)). So in this case, it's better if we simply keep the `fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.buildhub.json` part without its leading `latest ...` prefix.

beetmoverscript/templates/firefox_nightly.yml
```diff
@@ -129,6 +129,11 @@ mapping:
       destinations:
         - {{ upload_date }}-{{ branch }}/mozharness.zip
         - latest-{{ branch }}/mozharness.zip
+    buildhub.json:
+      s3_key: {{ upload_date }}-{{ branch }}/firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json
```
MihaiTabara: `s3_key` is solely used in checksums generation. Whatever we end up placing here, will be landing in the SHA512SUMS files (e.g. [here](https://archive.mozilla.org/pub/firefox/candidates/61.0b14-candidates/build1/SHA256SUMS)). So in this case, it's better if we simply keep the `firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json` part without its leading `upload_date ...` prefix.

beetmoverscript/templates/firefox_nightly_repacks.yml
```diff
@@ -30,6 +30,11 @@ mapping:
       destinations:
         - {{ upload_date }}-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.checksums.asc
         - latest-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.checksums.asc
+    buildhub.json:
+      s3_key: {{ upload_date }}-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json
```
MihaiTabara: `s3_key` is solely used in checksums generation. Whatever we end up placing here, will be landing in the SHA512SUMS files (e.g. [here](https://archive.mozilla.org/pub/firefox/candidates/61.0b14-candidates/build1/SHA256SUMS)). So in this case, it's better if we simply keep the `firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json` part without its leading `upload_date ...` prefix.

beetmoverscript/templates/firefox_partner.yml
```diff
@@ -12,4 +12,5 @@ mapping:
     target.installer.exe.asc: Firefox Setup {{ version }}.exe.asc
     target.dmg: Firefox {{ version }}.dmg
     target.dmg.asc: Firefox {{ version }}.dmg.asc
-    target.checksums: firefox-{{ version }}.checksums.beet
+    buildhub.json: buildhub.json
```
MihaiTabara: We don't have buildhub for partner templates. We can remove this.

beetmoverscript/templates/firefox_partner_repacks.yml
```diff
@@ -13,5 +13,6 @@ mapping:
     target.installer.exe.asc: Firefox Setup {{ version }}.exe.asc
     target.dmg: Firefox {{ version }}.dmg
     target.dmg.asc: Firefox {{ version }}.dmg.asc
+    buildhub.json: buildhub.json
```
MihaiTabara: We don't have buildhub for partner templates. We can remove this.

beetmoverscript/templates/thunderbird_candidates.yml
```diff
@@ -104,6 +104,10 @@ mapping:
       s3_key: beetmover-checksums/source/thunderbird-{{ version }}.checksums.asc
       destinations:
         - beetmover-checksums/source/thunderbird-{{ version }}.checksums.asc
+    buildhub.json:
```
MihaiTabara: This is harmless in the templates so we can definitely keep this. However, we'd need to double-check with @tomprince that `buildhub.json` is in the upstreamArtifacts for Thunderbird jobs. Otherwise, it won't update anything. Looking at the latest Thunderbird beta, sounds like it's not https://tools.taskcluster.net/groups/RTCkxBCKTYeozYKGn5ZYSg/tasks/JybFCvuJSFOmPRe-JBibLw/details but I strongly suspect the changes for buildhub.json were riding the trains so the newly generated betas on Thunderbird should have it. We need to talk to Tom for this. It'd be nice to fix this for Thunderbird as well since we're here.

beetmoverscript/test/test_script.py
```diff
@@ -516,6 +526,7 @@ def test_get_destination_for_partner_repack_path(context, full_path,
     context.release_props = get_release_props(context)
     mapping_manifest = generate_beetmover_manifest(context)
 
+    # import pdb; pdb.set_trace()
```
MihaiTabara: Let's remove this line. Even if its testing, debugging code should not make it in the final python package.

beetmoverscript/task.py
peterbe: +1 on `urllib.parse.urljoin` 
Equally, no need to use `os.path.join` or `url.split('/')`.

lisajguo: got it, hopefully it's good this time 😅

MihaiTabara:
https://tools.taskcluster.net/groups/DEmurrqBRraV93XPOPpNWQ is the latest staging release based on tip of this PR.

MihaiTabara:
Triggering another staging release https://tools.taskcluster.net/groups/AJG4k7djR-eKSLQsP4G-FQ as the previous one exceeded the deadline for a bunch of tasks for whatever reasons, most likely low-capacity.  Let's see how this one goes

beetmoverscript/templates/fennec_nightly.yml
```diff
@@ -128,5 +128,9 @@ mapping:
       destinations:
         - latest-{{ branch }}-{{ platform }}/{{ locale_prefix }}robocop.apk
         - {{ upload_date }}-{{ branch }}-{{ platform }}/{{ locale_prefix }}robocop.apk
-
+    buildhub.json:
+      s3_key: fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.buildhub.json
+      destinations:
+        - latest-{{ branch }}-{{ platform }}/{{ locale_prefix }}.buildhub.json
```
MihaiTabara: Let's stick to `fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.buildhub.json` to for these filenames. AFAIK the `{{ locale_prefix }}` evaluates to empty string for `en-US` so that file will be slightly confusing, at least in the `latest...` dirs. I'd rather opt for 

```
latest-{{ branch }}-{{ platform }}/fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.buildhub.json
{{ upload_date }}-{{ branch }}-{{ platform }}/fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.buildhub.json
```

beetmoverscript/templates/fennec_nightly_repacks.yml
```diff
@@ -25,4 +25,9 @@ mapping:
       destinations:
         - {{ upload_date }}-{{ branch }}-{{ platform }}-l10n/fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.checksums.asc
         - latest-{{ branch }}-{{ platform }}-l10n/fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.checksums.asc
+    buildhub.json:
+      s3_key: fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.buildhub.json
+      destinations:
+        - {{ upload_date }}-{{ branch }}-{{ platform }}-l10n/fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.buildhub.json
+        - latest-{{ branch }}-{{ platform }}-l10n/fennec-{{ version }}.{{ locale }}.{{ filename_platform }}.buildhub.json
```
MihaiTabara: Very good!

beetmoverscript/templates/firefox_nightly.yml
MihaiTabara: This is still wrong. It should be `firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json` instead

beetmoverscript/templates/firefox_nightly_repacks.yml
```diff
@@ -30,6 +30,11 @@ mapping:
       destinations:
         - {{ upload_date }}-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.checksums.asc
         - latest-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.checksums.asc
+    buildhub.json:
+      s3_key: firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json
+      destinations:
+        - {{ upload_date }}-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json
+        - latest-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json
```
MihaiTabara: Very good!

beetmoverscript/templates/thunderbird_candidates.yml
```diff
@@ -104,6 +104,10 @@ mapping:
...
+      s3_key: {{ platform }}/{{ locale }}.buildhub.json
```
MihaiTabara: I assume you meant `{{ platform }}/{{ locale }}.buildhub.json` here instead.

beetmoverscript/templates/thunderbird_candidates.yml
```diff
@@ -104,6 +104,10 @@ mapping:
...
+      destinations:
+        - {{ platform }}/{{ locale }}.buildhub.json
```
MihaiTabara: I assume you meant `{{ platform }}/{{ locale }}.buildhub.json` here instead.

beetmoverscript/templates/thunderbird_candidates_repacks.yml
```diff
@@ -18,6 +18,10 @@ mapping:
       s3_key: {{ platform }}/xpi/{{ locale }}.xpi
       destinations:
         - {{ platform }}/xpi/{{ locale }}.xpi
+    buildhub.json:
+      s3_key: {{ platform }}/{{ locale }}.buildhub.json
+      destinations:
+        - {{ platform }}/{{ locale }}.buildhub.json
```
MihaiTabara: I assume you meant `{{ platform }}/{{ locale }}.buildhub.json` here instead in both occurrences.

beetmoverscript/templates/thunderbird_candidates.yml
lisajguo: this is what I have, no?

beetmoverscript/templates/thunderbird_candidates_repacks.yml
lisajguo: I believe I have that?

beetmoverscript/templates/firefox_nightly.yml
```diff
@@ -129,6 +129,12 @@ mapping:
       destinations:
         - {{ upload_date }}-{{ branch }}/mozharness.zip
         - latest-{{ branch }}/mozharness.zip
+    buildhub.json:
+      s3_key: firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json
+      destinations:
+        - {{ upload_date }}-{{ branch }}/firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json
+        - latest-{{ branch }}/firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json
+        - latest-{{ branch }}-l10n/firefox-{{ version }}.{{ locale }}.{{ platform }}.buildhub.json
```
MihaiTabara: Not sure why we have three entrie for "firefox_nightly" but not the same for "fennec_nightly". However, I think it's useless. As far as I can tell from Peter's [comments](https://bugzilla.mozilla.org/show_bug.cgi?id=1443873#c16)  they are going to consume the artifacts as they are, from the entire dir structure / release so I don't think it's a problem.

MihaiTabara:
Devedition - https://tools.taskcluster.net/groups/NMQWT4ekQtmLjLfdRTX1UA
Fennec - https://tools.taskcluster.net/groups/FF9Yhi-tQ0CHpAG588i66Q
Firefox - https://tools.taskcluster.net/groups/BBNtJvQPTdmZv2aVu5sY0A

If all these three staging releases are fine, I think we can land this to production.

MihaiTabara:
All three staging releases look good. From my end, we're ready to land this to production, once 61.0 is out of the door later on today.

MihaiTabara:
@lisajguo   FYI - this is getting to production in a bit via [7.5.0](https://github.com/mozilla-releng/beetmoverscript/commit/dbaf1471262e487463fdc83fb7d153053198e430) roll-out.

PR #156, opened 2018-06-07
Improve README with towncrier instructions.

MihaiTabara:


README.md
```diff
@@ -54,15 +54,17 @@ sudo puppet agent --test  # or unpin or w/e
 
 1. Once your PR is reviewed and passes the tests, have one of the admins review & merge it
 1. Bump to new version in `beetmoverscript/_version.py`.
-1. `towncrier`
+1. Sum-up the changes in the `beetmoverscript/newsfragments/` in the form of `bug_no.added` or `bug_no.fixed` or `bug_no.changed` or `bug_no.removed` files. More on this can be found [here](https://pypi.org/project/towncrier/)
```
tomprince: This should really be something that is done as part of each PR, rather than requiring somebody to go through all the changes at release time. Maybe it would make sense to add a PR template that has a checkbox for adding a news fragment instead?

MihaiTabara: At second glance, I think you're right and we shouldn't go with these changes forward. Each PR should have its own newsfragment section. 

Having a PR template - I think that's not a bad idea. We're a handful of people that approve these PRs anyway so we can start pointing to that best practice until it becomes habit. I'll update the PR in a bit.

MihaiTabara:
Ditching this PR as README fixes were picked up by @nthomas-mozilla in https://github.com/mozilla-releng/beetmoverscript/pull/158 PR. I'll follow-up with a separate PR to take into account what @tomprince suggested - https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/

PR #157, opened 2018-06-08
Attempt to fix coveralls template errors.

MihaiTabara:


tomprince:
I don't think those templates are incorrect. I think coveralls is incorrectly trying to parse yaml as python.

`---` is yaml's document start syntax

MihaiTabara:
Yeah, you're totally right. I just thought it's worth a blind shot. I'll investigate more

PR #163, opened 2018-06-25
Bug 1470942 - Support maven on S3

JohanLorenzo:
~~**Do not merge yet**. This has been untested from an end to another. This is just to let people know this kind of changes are incoming. In my opinion, they aren't too heavy and shouldn't (hopefully) block any ongoing work. Please let me know if that's not the case.~~ All done!


beetmoverscript/zip.py
```diff
@@ -0,0 +1,206 @@
...
+        raise TaskVerificationError(
+            'Found duplicated expected files in archive "{}": {}'.format(zip_path, duplicated_files)
+        )
+
+    for file in files_in_archive:
+        if os.path.isabs(file):
+            raise TaskVerificationError(
+                'File "{}" in archive "{}" cannot be an absolute one.'.format(file, zip_path)
+            )
+        if os.path.normpath(file) != file:
```
MihaiTabara: Smart check, nice one! 🥇

beetmoverscript/zip.py
```diff
@@ -0,0 +1,206 @@
...
+
+    with zipfile.ZipFile(zip_path) as zip_file:
+        zip_metadata = _fetch_zip_metadata(zip_file)
+
+        # we don't close the file descriptor here to avoid the tested file to be swapped by a rogue one
+        _ensure_files_in_archive_have_decent_sizes(zip_path, zip_metadata, zip_max_size_in_mb)
+        _ensure_all_expected_files_are_present_in_archive(zip_path, zip_metadata, expected_files)
+        log.info('Content of archive "{}" is sane'.format(zip_path))
+
+        extracted_files = _extract_and_check_output_files(zip_file, zip_metadata.keys())
```
MihaiTabara: Nit: could be my own preferences so feel free to ignore this. Usually I avoid having magic logic happening in the params. I'd rather submit zip_metadata as dict here, and do that processing with keys extraction in the function itself.

beetmoverscript/zip.py
```diff
@@ -0,0 +1,206 @@
...
+    if len(files_in_archive) != len(unique_expected_files):
+        missing_expected_files = [file for file in unique_expected_files if file not in files_in_archive]
+        raise TaskVerificationError(
+            'Expected files are missing in archive "{}": {}'.format(zip_path, missing_expected_files)
+        )
+
+    log.info('Archive "{}" contains all expected files: {}'.format(zip_path, unique_expected_files))
+
+
+def _extract_and_check_output_files(zip_file, expected_files_in_archive):
```
MihaiTabara: Similarly as one of the functions above, I'd rather keep the `s/expected_files_in_arhive/zip_metadata` and do that `zip_metadata.keys()` processing in this function to enhance the logic reading. But it's a matter of preference, I don't have a strong feeling nor way or another.

beetmoverscript/utils.py
```diff
@@ -81,6 +81,13 @@ def is_partner_action(action):
     return action in PARTNER_REPACK_ACTIONS
 
 
+def is_maven_action(action):
+    """Function to return boolean if we're promoting a release as opposed to a
```
MihaiTabara: Hm, I feel like the comment could be enhanced a bit by adding something maven or Geckoview specific like: "we're promoting a release with gecko view components as opposed to a nightly or regular release" or alike. What do you think? Afterall, it's a total separate behavior, even though the behind-the-scenes S3-mechanics are the reused.

beetmoverscript/utils.py
```diff
@@ -161,6 +171,26 @@ def generate_beetmover_template_args(context):
...
+        'artifact_id': task['payload']['artifact_id'],
+        'branch': release_props['branch'],
+        'buildid': release_props['buildid'],
+        'product': release_props['appName'],
+        'template_key': 'maven_{}'.format(release_props['appName']),
+        'version': task['payload']['version'],
+    }
+
+    if 'nightly' in tmpl_args['artifact_id']:
+        # Change version number to major.minor.buildId
```
MihaiTabara: ++++ to using mozilla-version here!

beetmoverscript/utils.py
```diff
@@ -161,6 +171,26 @@ def generate_beetmover_template_args(context):
...
+    tmpl_args = {
+        'artifact_id': task['payload']['artifact_id'],
+        'branch': release_props['branch'],
+        'buildid': release_props['buildid'],
+        'product': release_props['appName'],
+        'template_key': 'maven_{}'.format(release_props['appName']),
+        'version': task['payload']['version'],
+    }
+
+    if 'nightly' in tmpl_args['artifact_id']:
```
MihaiTabara: I'm confused. Can you **not** have `nightly` there as a string? I'm looking at https://tools.taskcluster.net/groups/TJW4b-CNSwGsJukgQCeoUA/tasks/PojKrc74Q5KFqghgPlosCg/details 🤔 

... later edit: never mind me. I found https://github.com/mozilla-releng/beetmoverscript/pull/163/files#diff-8a0baf3b191b3039ec72b8da4e543908R490

config_example.json
```diff
@@ -56,5 +56,7 @@
                 "firefox": "mozilla-releng-private-firefox-bucket"
             },
         }
-    }
+    },
+
+    "zip_extract_max_file_size_in_mb": 100,
```
MihaiTabara: I like it better here than in constants.py 👍

beetmoverscript/test/test_zip.py
```diff
@@ -0,0 +1,418 @@
+import contextlib
```
MihaiTabara: Needless to say this is simply beautiful!

beetmoverscript/test/test_utils.py
```diff
@@ -218,6 +487,32 @@ def test_beetmover_template_args_generation_release(context):
     assert template_args == expected_template_args
 
 
+@pytest.mark.parametrize('branch, version, artifact_id, build_id, expected_version', ((
+    'mozilla-central', '63.0a1', 'geckoview-nightly-x86', '20181231120000', '63.0.20181231120000',
+), (
+    'mozilla-beta', '63.0b2', 'geckoview-beta-armeabi-v7a', '20181231120000', '63.0b2',
+), (
+    'mozilla-release', '63.0', 'geckoview-arm64-v8a', '20181231120000', '63.0',
+)))
```
MihaiTabara: nit: might be worth adding something for release/esr as well here, just for the sake of consistency?

beetmoverscript/templates/maven_geckoview.yml
```diff
@@ -0,0 +1,62 @@
...
+metadata:
+    name: "Maven repository"
+    description: "Maps artifacts to spec'd maven location"
+    owner: "release@mozilla.com"
+
+s3_bucket_path: maven2/org/mozilla/{{ artifact_id }}/{{ version }}/  # Maven groupId is org.mozilla
+
+mapping:
+{% for locale in ['en-US'] %}
+  "{{ locale }}":   # Locale is not needed for geckoview, it's used by move_beets, though
```
MihaiTabara: Yeah, I feel you. However this is going to be washed away when declarative artifacts is done.

beetmoverscript/templates/maven_geckoview.yml
```diff
@@ -0,0 +1,62 @@
...
+{% for product in ['geckoview'] %}
```
MihaiTabara: Tiny nit: can we add a secondary tab here to indent properly the manifest so that's easier to see it's a nested `for`?

beetmoverscript/task.py
```diff
@@ -109,29 +110,45 @@ def add_balrog_manifest_to_artifacts(context):
     utils.write_json(abs_file_path, context.balrog_manifest)
 
 
-def get_upstream_artifact(context, taskid, path):
```
MihaiTabara: Hah, nice cleanup, didn't know of `scriptworker_artifacts.get_and_check_single_upstream_artifact_full_path` 👍

beetmoverscript/script.py
```diff
@@ -154,6 +156,43 @@
...
+        context.artifacts_to_beetmove, mapping_manifest
+    )
+
+    extracted_paths_per_archive = zip.check_and_extract_zip_archives(
+        context.artifacts_to_beetmove,
+        expected_files,
+        context.config.get('zip_extract_max_file_size_in_mb', ZIP_MAX_FILE_SIZE_IN_MB)
+    )
+
+    number_of_extracted_archives = len(extracted_paths_per_archive)
```
MihaiTabara: I somehow feel like these sanity checks could be moved elsewhere in a separate function but I may be wrong. The other `push_to_*` methods largely avoided doing any checks (well, frankly there are some in`push_to_releases` IIRC). I have no strong feelings though and this isn't blocking. 

It's just that these functions, the way I see them, are blocks of logic to follow-up what's going on. But the actual logic, validations, checks and processing happens behind the scenes. What do you think?

beetmoverscript/script.py
```diff
@@ -154,6 +156,43 @@
...
+    if number_of_extracted_archives == 0:
+        raise ScriptWorkerTaskException('No archive extracted')
+    elif number_of_extracted_archives > 1:
+        raise NotImplementedError('More than 1 archive extracted. Only 1 is supported at once')
+    extracted_paths_per_relative_path = list(extracted_paths_per_archive.values())[0]
+
+    context.artifacts_to_beetmove = {
```
MihaiTabara: I was wondering if it's worth moving this `context.artifacts_to_beetmove` logic + logic from 169-179 in a separate function. Again, no strong feelings, we can leave them as they are, I'm more leaning towards consistency with the other `push_to_*` operations.

beetmoverscript/maven.py
```diff
@@ -0,0 +1,54 @@
+import os
```
MihaiTabara: I'm tempted to see this file renamed to `maven_utils.py`. Not sure if it's worth it though. What say you?

beetmoverscript/data/release_beetmover_task_schema.json
```diff
@@ -46,9 +46,12 @@
                                 "items": {
                                     "type": "string"
                                 }
+                            },
+                            "zipExtract": {
+                                "type": "boolean"
                             }
                         },
-                        "required": ["taskId", "taskType", "paths", "locale"]
```
MihaiTabara: Why trim `locale` here?

beetmoverscript/data/release_beetmover_task_schema.json
```diff
@@ -46,9 +46,12 @@
                                 "items": {
                                     "type": "string"
                                 }
+                            },
```
MihaiTabara: Note to self: in the near future, in my backlog, I have a task to add integration tests for beetmover in such a way that we have beetmover schema for all possible beetmover jobs + actual tests against them. Like we currently have in bouncerscript or shipitscript. Right now we are reusing these two possible tasks but it's not enough.

beetmoverscript/data/beetmover_task_schema.json
```diff
@@ -79,9 +79,12 @@
                                 "items": {
                                     "type": "string"
                                 }
+                            },
+                            "zipExtract": {
+                                "type": "boolean"
                             }
                         },
-                        "required": ["taskId", "taskType", "paths", "locale"]
```
MihaiTabara: I presume you remove 'locale' to fit the maven jobs right? The old beetmover jobs are still having the `locale`. Maybe it'd be worth adding it as optional.

beetmoverscript/constants.py
```diff
@@ -163,3 +173,9 @@
     'target.dmg',
     'target.apk',
 )
+
+# Zip archive can theoretically have a better compression ratio, like when there's a big amount
```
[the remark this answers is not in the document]
JohanLorenzo: Oh good call. I had forgotten about script_config. I added the max file size in https://github.com/mozilla-releng/build-puppet/pull/87/commits/bcef5fc6d4385f6b3d539c641b99393262e4b5bb.

I don't foresee a need to change the compression ratio as much as the max file size. So, I haven't added it. Moreover, I think increasing the ratio can be a bigger risk than just increasing the file size. If one day we realize we need to deal with zip files that are compressed more than 10 times, we probably want to revisit the ratio check within beetmover. That's why, I think we shouldn't expose that value in script_config.

beetmoverscript/constants.py
```diff
@@ -163,3 +173,9 @@
...
+# of redundancy (e.g.: files full of zeros). Let beetmover only deal with regular cases. Edge cases
+# are considered too suspicious, so we bail out on them.
+ZIP_MAX_FILE_SIZE_IN_MB = 100
```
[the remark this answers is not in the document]
JohanLorenzo: Understood. I see the confusion. How about I rename the constant ZIP_MAX_FILE_SIZE_IN_MB into DEFAULT_ZIP_MAX_FILE_SIZE_IN_MB. This should make the intent of this constant clearer. I made this change in https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/42ade2a4b65f50d17a3932423e2d055ba79d547a.

Please let me know what you think of it!

beetmoverscript/zip.py
```diff
@@ -0,0 +1,206 @@
+import logging
```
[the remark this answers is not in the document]
JohanLorenzo: Okay. I added the doctstring of the only public function in https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/3c02cc551c5a947dfd1104405c3e7561bd3d0318. I usually prefer to not document private functions in favor of having (long) but explicit names. The reason is: this is an implementation detail to the consumers of the module. Ergo, consumers should not rely on these functions but documenting them can send mixed signals.

I sometimes briefly document private functions (like https://github.com/mozilla-releng/mozilla-version/blob/master/mozilla_version/firefox.py#L245), if the function name cannot capture the intent and the function is too obscure to guess the output. 

Please let me know if you still want me to document these private functions.

beetmoverscript/zip.py
```diff
@@ -0,0 +1,206 @@
...
+
+    for archive_path, expected_files in expected_files_per_archive.items():
+        log.info('Processing archive "{}" which marked as `zipExtract`able'.format(archive_path))
+        extracted_files[archive_path] = _check_extract_and_delete_zip_archive(
+            archive_path, expected_files, zip_max_size_in_mb
+        )
+
+    # We make this check at this stage (and not when all files from all tasks got extracted)
+    # because files from different tasks are stored in different folders by scriptworker. Moreover
+    # we tested no relative paths like ".." are not used within the archive.
```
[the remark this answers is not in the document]
JohanLorenzo: Thanks! Done in: https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/b7470991deed76b10ba1f74a194d032a54e33759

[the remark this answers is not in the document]
JohanLorenzo: More precisely here https://github.com/mozilla-releng/beetmoverscript/pull/163/files#diff-588c919d6a181374d2dcebb61ff5fe39R174 https://docs.python.org/3/library/os.path.html#os.path.normpath

beetmoverscript/zip.py
```diff
@@ -0,0 +1,206 @@
...
+    # We remove the zip archive because it's not used anymore. We just need the deflated files
+    os.remove(zip_path)
+    log.debug('Deleted archive "{}"'.format(zip_path))
+
+    return extracted_files
+
+
+def _check_archive_itself(zip_path, zip_max_size_in_mb):
+    zip_size = os.path.getsize(zip_path)
+    zip_size_in_mb = zip_size // (1024 * 1024)
```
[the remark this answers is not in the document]
JohanLorenzo: I agree. If this use case comes up again, let's make a function for it.

beetmoverscript/zip.py
```diff
@@ -0,0 +1,206 @@
...
+            raise TaskVerificationError(
+                'In archive "{}", file "{}" has a suspicious compression ratio. Max accepted: {}. Found: {}'.format(
+                    zip_path, file_name, ZIP_MAX_COMPRESSION_RATIO, compression_ratio
+                )
+            )
+
+    log.info('Archive "{}" contains files with legitimate sizes.'.format(zip_path))
+
+
+def _ensure_all_expected_files_are_present_in_archive(zip_path, files_in_archive, expected_files):
```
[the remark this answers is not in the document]
JohanLorenzo: Great point! Fixed in https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/5d9fcf0caccb059410bfc7ddca9ecefae1a0cdb2

beetmoverscript/zip.py
```diff
@@ -0,0 +1,206 @@
...
+    files_in_archive = set(files_in_archive)
+
+    unique_expected_files = set(expected_files)
+    if len(expected_files) != len(unique_expected_files):
```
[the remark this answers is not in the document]
JohanLorenzo: That's right. Here, we check the list before we extract it. At the other place, we verify if what got extracted is still matching. I'm not aware of a case where a zip archive lies about the file list in its metadata, but it seems safer to eliminate the archives before extracting anything and then ensuring we still have the files we checked earlier.

beetmoverscript/zip.py
```diff
@@ -0,0 +1,206 @@
...
+        duplicated_files = [file for file in unique_expected_files if expected_files.count(file) > 1]
+        raise TaskVerificationError(
+            'Found duplicated expected files in archive "{}": {}'.format(zip_path, duplicated_files)
+        )
+
+    for file in files_in_archive:
```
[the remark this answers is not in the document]
JohanLorenzo: Oops, you're right! Done

beetmoverscript/zip.py
JohanLorenzo: Yeah, I agree this is confusing. Thank you for spotting it. I fixed this in https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/5d9fcf0caccb059410bfc7ddca9ecefae1a0cdb2 and https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/86f9770b59d509a3ad11871354248b165b9122c3. Please tell me what you think!

beetmoverscript/zip.py
JohanLorenzo: Done in https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/86f9770b59d509a3ad11871354248b165b9122c3

beetmoverscript/utils.py
JohanLorenzo: Oh right, I think I copied this piece of doc and forgot to update it. Done in https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/6a0a75dc31baa88bb7bef9103abe7f4fdba4df23. Thank you for spotting it!

beetmoverscript/utils.py
JohanLorenzo: Yeah, nightly occurs on maple and mozilla-central. `artifact_id` is gotten from _MOZ_UPDATE_CHANNEL_PER_BRANCH in https://phabricator.services.mozilla.com/D2421 which maps what the build system creates. 

For the record, I didn't manage to get the config file from the build system in-tree. I just found a mozconfig on maple: https://searchfox.org/mozilla-central/rev/033d45ca70ff32acf04286244644d19308c359d5/testing/mozharness/configs/single_locale/maple_android-api-16.py#28

beetmoverscript/test/test_zip.py
JohanLorenzo: Thank you :)

beetmoverscript/test/test_utils.py
JohanLorenzo: Right. For the record, ESR is not supported as geckoview is an android build and Android deactivated on ESR..

beetmoverscript/templates/maven_geckoview.yml
JohanLorenzo: Sure. Done in https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/bb88611b5a78aecfcb45f7ea4c897946bd485718

beetmoverscript/task.py
JohanLorenzo: I just realize I implemented it a while ago in https://github.com/mozilla-releng/scriptworker/pull/95. I knew beetmover had its own logic, but I didn't cc you on it. Sorry about this

beetmoverscript/maven.py
JohanLorenzo: Fine by me. Done in https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/5cab293b61869aaba9bfdb5c5ad756467a09bccd

beetmoverscript/data/release_beetmover_task_schema.json
JohanLorenzo: geckoview tasks don't specify a locale to beetmove. That's why I removed it from the required fields. I'm fine putting it back and creating a different schema. What do you think?

beetmoverscript/data/beetmover_task_schema.json
JohanLorenzo: Yup, that's what this change is about. `locale` is now optional

beetmoverscript/script.py
JohanLorenzo: Yeah, I was wondering the same. I can't find a name for this function. I get the feeling too many concepts are intertwined. That's why I left it in this function. How would you describe this function?

JohanLorenzo:
Addressed comments were tested in https://tools.taskcluster.net/groups/NaetTOqNQqOLoI0cGDH2_Q/tasks/JDuBbPHpRaqAvOH4mR2JXA/runs/0/logs/public%2Flogs%2Flive_backing.log#L71

beetmoverscript/constants.py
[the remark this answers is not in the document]
MihaiTabara: That makes sense, thanks for clarifying.

beetmoverscript/data/beetmover_task_schema.json
MihaiTabara: 👍

beetmoverscript/data/release_beetmover_task_schema.json
MihaiTabara: This is my fault from the very beginnings of beetmoverscript. In `shipitscript` and `bouncerscript` we have task schemas for each of the possible tasks and full integration tests as we should have! For beetmover unfortunately it is what it is. So to answer your question, ideally, we add a new schema so that we don't touch existing ones (like trimming the `locale` or `hashType` or `platform` to fit new maven tasks). Old beetmover jobs still have those, hence be tested against those correctly.

Again, this is not blocking so we can definitely follow-up in a separate PR. I have a backlog task to take over (hopefully) within declarative artifacts, to rewrite all the tasks to have integration tests for all possible beetmover jobs so this work will be undertaken anyway at some point this quarter so not sure it's worth time investing now. If it's easy for you to tweak tasks to fit to a new schema and not touch existing, that's be ideal, but again, we're going to do that sooner or later anyway for all tasks. Up to you, am fine either way 👍

beetmoverscript/script.py
MihaiTabara: Hm, good question. I wonder, how bad it would be if we defined that function within push_to_maven()? Something like:

```
async def push_to_maven(context):
       def _process_and_validate_artifacts(...)
        ....
        return artifacts
        
        ......
       context.artifacts_to_beetmove = _process_and_validate_artifacts(...)
```

I know most of the other similar functions are cleaner but if we don't find any better solution, this could be a good compromise. Alternatively, we can leave things be, I don't want to block on this, nor I have strong feelings.

beetmoverscript/task.py
MihaiTabara: No, absolutely no worries, this is super nice! 👍

beetmoverscript/utils.py
MihaiTabara: Hm, I feel you, would've been better if it came from in-tree I suppose. This might be fragile in the future but sounds like it works for now; The only place I know that var is as well is [this](https://hg.mozilla.org/mozilla-central/file/tip/browser/confvars.sh) but I don't think that helps much.

beetmoverscript/zip.py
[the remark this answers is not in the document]
MihaiTabara: Public is perfect; your argument for private ones makes a lot of sense. Now I understand retroactively why you were always suggesting the best of the names when reviewing 😃 👍

beetmoverscript/zip.py
```diff
@@ -0,0 +1,244 @@
...
+
+from beetmoverscript.constants import ZIP_MAX_COMPRESSION_RATIO
+
+log = logging.getLogger(__name__)
+
+
+def check_and_extract_zip_archives(artifacts_per_task_id, expected_files_per_archive_per_task_id, zip_max_size_in_mb):
+    """Verify zip archives and extract them.
+
+    This function enhances the checks done in python's zipfile. Each of the zip file passed is
```
MihaiTabara: Holy smokes, this is docs for realz 😮

beetmoverscript/zip.py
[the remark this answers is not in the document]
MihaiTabara: 👍

beetmoverscript/zip.py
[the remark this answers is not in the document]
MihaiTabara: \o/

beetmoverscript/zip.py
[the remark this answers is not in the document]
MihaiTabara: Sounds good to me, thanks for adding all these checks. Makes me feel so much safer having all of them. GG again for all the validations!

beetmoverscript/data/release_beetmover_task_schema.json
JohanLorenzo: Got! I'd prefer to make it in a follow up. I think we're safe for now because beetmoverscript relies on `locale` being defined. So if a task def misses one, a `KeyError` will be raised somewhere

beetmoverscript/script.py
JohanLorenzo: Done in https://github.com/mozilla-releng/beetmoverscript/pull/163/commits/7f5682ee62a71f78ea11e1e8ac4b4550d713f409. Unit tests are failing because of #176.

Tested e2e in https://tools.taskcluster.net/groups/FN8cz6cTSyiKeVxa8Y0KAg

MihaiTabara: Looks great, thanks for making this amendment!

beetmoverscript/data/release_beetmover_task_schema.json
MihaiTabara: 👍

PR #167, opened 2018-07-04
declarative artifacts design collab

lisajguo:


beetmoverscript/templates/mini_artifact_manifest.yml
```diff
@@ -0,0 +1,603 @@
...
+  thunderbird_nightly_repacks: pub/thunderbird/nightly/
+mapping:
+  win32:
+    partials[partial].artifact_name:
+      default: partials[partial].artifact_name
+      devedition:
+        candidates:
+          pretty_name: update/{{ platform }}/{{ locale }}/firefox-{{ partials[partial].previousVersion }}-{{ version }}.partial.mar
+          pretty_path:
+          - update/{{ platform }}/{{ locale }}/firefox-{{ partials[partial].previousVersion }}-{{ version }}.partial.mar
```
escapewindow: I know I made some of these suggestions initially, but now that it's populated, it looks like the pretty name and pretty path contain the exact same info, which leads me to the questions/observations:

1. do we gain anything by having a `pretty_name` defined if it's duplicated in `pretty_path`?
2. if we do gain something, should the `pretty_name` be the filename without any path info?

I'm also torn whether we'll ever use multiple `pretty_path`s (YAGNI) or if we'll be glad we have support for it for certain files. I bet we'll gather more opinions about this when we send this out more widely.

beetmoverscript/templates/mini_artifact_manifest.yml
```diff
@@ -0,0 +1,603 @@
...
+  firefox_candidates_repacks: pub/firefox/candidates/{{ version }}-candidates/build{{ build_number }}/
+  firefox_nightly: pub/firefox/nightly/
+  firefox_nightly_repacks: pub/firefox/nightly/
+  thunderbird_candidates: pub/thunderbird/candidates/{{ version }}-candidates/build{{ build_number }}/
+  thunderbird_candidates_repacks: pub/thunderbird/candidates/{{ version }}-candidates/build{{ build_number }}/
+  thunderbird_nightly: pub/thunderbird/nightly/
+  thunderbird_nightly_repacks: pub/thunderbird/nightly/
+mapping:
+  win32:
+    partials[partial].artifact_name:
```
escapewindow: It looks like your current artifact keys are the simple name. I think this might work, especially for people who are already familiar with the artifacts. However, something like `target.json` or `target.txt` isn't especially descriptive, and something like `version-info-json` may help. Alternately, adding a `description` field may help, so we can say

```
target.json:
    description: a json file containing version information for releases
    default: ...
    ...
```

Another issue with having the filename match the key is that it hardcodes the extension. This might be fine for json or txt, assuming we don't ever want to change the extension. Catlee is currently changing the format for the test `zip`s to `tar.gz`, and may change the extension/format once again. Since it's all in-tree, we can do a recursive find/replace... I'm wondering if it makes things easier or harder if we do something like

```
common-tests-bundle:  # I'm not in love with this name, just the first thing that came to mind
    description: Files common to the various browser tests
    simple: target.common.tests.zip
    nightly: ...
    ...
```

instead of

```
target.common.tests.zip:
    description: ...
```

That assumes we'd have functions that know how to create and extract various extensions, so maybe it is better to replace everything. Not sure.

beetmoverscript/templates/mini_artifact_manifest.yml
```diff
@@ -0,0 +1,603 @@
...
+  owner: release@mozilla.com
+s3_bucket_paths:
+  devedition_candidates: pub/devedition/candidates/{{ version }}-candidates/build{{ build_number }}/
+  devedition_candidates_repacks: pub/devedition/candidates/{{ version }}-candidates/build{{ build_number }}/
+  fennec_candidates_repacks: pub/mobile/candidates/{{ version }}-candidates/build{{ build_number }}/
+  fennec_nightly: pub/mobile/nightly/
+  fennec_nightly_repacks: pub/mobile/nightly/
+  firefox_candidates: pub/firefox/candidates/{{ version }}-candidates/build{{ build_number }}/
+  firefox_candidates_repacks: pub/firefox/candidates/{{ version }}-candidates/build{{ build_number }}/
+  firefox_nightly: pub/firefox/nightly/
```
escapewindow: I think splitting these bucket paths out is smart in that it avoids repeating these strings throughout the file. I'm wondering if we also want to put in things like nightly dated dir paths here, or in an `s3_subdirs` section.

beetmoverscript/templates/mini_artifact_manifest.yml
```diff
@@ -0,0 +1,603 @@
...
+        nightly_repacks:
+          pretty_name: ''
+          pretty_path: ''
+        partner:
+          pretty_name: ''
+          pretty_path: ''
+        partner_repacks:
+          pretty_name: ''
+          pretty_path: ''
+      simple: partials[partial].artifact_name
```
escapewindow: If `simple` never deviates from `default`, we could drop one.

lisajguo: So far, this is just how I've designed it because I don't know how the Build team might want to use default - it may change

beetmoverscript/templates/mini_artifact_manifest.yml
MihaiTabara: Hm, I think `s3_subdirs` is a good idea. Especially because having `pretty_path`s (hence a list instead of a value) for some of the entries might be confusing even more and look ugly. 

If we were to choose `target.crashreporter-symbols.zip:` (e.g. from [here](https://github.com/mozilla-releng/beetmoverscript/blob/master/beetmoverscript/templates/firefox_nightly.yml#L62), the way I'd see the manifest for it would be something like: 
```
    target.crashreporter-symbols.zip:
      default: target.crashreporter-symbols.zip
      devedition:
        candidates:
          pretty_name: firefox-{{ version }}.crashreporter-symbols.zip
          pretty_path: {{ platform }}/{{ locale }}/firefox-{{ version }}.crashreporter-symbols.zip
...
        nightly:
          pretty_name: firefox-{{ version }}.{{ locale }}.{{ platform }}.crashreporter-symbols-full.zip
          pretty_path: ''
          s3_subdirs: 
               - latest-{{ branch }}/
               - {{ upload_date }}-{{ branch }}/
```

`pretty_path` would stay empty in this case. 
I see full path of an artifact being constructed like:

http://archive.mozilla.org/ + `s3_bucket_paths` + `s3_subdirs`(if any) + `pretty_path` + `pretty_name`

This assumes that the `pretty_name` +`pretty_path` stay the same if we have `s3_subdirs` which I'm not 100% sure is the case for all artifacts. I can't think a counter-example but I'm wondering if that's possible.

escapewindow:
We came up with 3 new concerns:

- special artifacts, e.g. target.dmg (build) -> target.tar.gz (signing) -> target.dmg (repackage)
   - `target.extension` might be used in multiple platforms, and end up with different end names. having a nickname may help here
- can the manifest specify which tasks use which artifacts? or will we need a file in taskgraph specifying still?
   - the [constants in beetmover*.py transforms](https://searchfox.org/mozilla-central/source/taskcluster/taskgraph/transforms/beetmover.py#26) -- we probably want to move these to the manifest, or a second file.
- do we assume `public/build` all the time, or will we want that in the manifest? this may be important for e.g. partners where it's private

@MihaiTabara and I brainstormed expanding all paths in the transforms, and then giving beetmoverscript a list of non-templatized pretty paths per artifact, e.g.

```
    "artifactMap": {
      "<task_id>" : {
        "public/build/target.tar.bz2": {
            "pretty_paths": [
               "pub/firefox/nightly/2018/07/05/linux/firefox-63.0a1.tar.bz2",
            ]
          }
       }
    }
```

If we don't foresee needing to add any more info, we could make it

```
"public/build/target.tar.bz2": [pretty_path1, pretty_path2]
```

rather than a dictionary. However, a dictionary allows for adding more info in the future.

beetmoverscript/templates/mini_artifact_manifest.yml
lisajguo: pretty_name = s3_key
pretty_path = destinations
... in the current yml templates.
I can change it back to "s3_key" and "destinations" if you want!

lisajguo: Mihai would know best but I think you're right Aki, that we could do without pretty_name/s3_key if it's just the basename of pretty_path/destinations

lisajguo:
Mihai and I just talked about partials:

we should name partials in the manifest something like "target.partial.mar" as a generic way to identify them in the manifest without logic. In the simple name, we can specify:  `target-{{ version }}-partial.mar`

escapewindow:
> we should name partials in the manifest something like "target.partial.mar" as a generic way to identify them in the manifest without logic. In the simple name, we can specify: target-{{ version }}-partial.mar

I think `{{ version }}` refers to the current version. In the partial artifact, the version is the previous version. We may want to use `previous_version` or something. We currently have multiples of these per partial task, e.g. https://tools.taskcluster.net/groups/Ei6VY3A1S_6b8_eT7wvbAA/tasks/A7CfEsUfRPiFZdnsosCf5Q/runs/0/artifacts has `public/build/af/target-62.0b6.partial.mar`, `public/build/af/target-62.0b7.partial.mar`, and `public/build/af/target-62.0b8.partial.mar`

PR #169, opened 2018-07-04
stop using scriptworker event_loop fixture

escapewindow:
This fixture is being removed in mozilla-releng/scriptworker#241 ; let's stop
using it.


tox.ini
```diff
@@ -42,6 +42,6 @@ exclude = .ropeproject,.tox,sandbox
 show-source = True
 
 [pytest]
-norecursedirs = .tox .git .hg sandbox
+norecursedirs = .tox .git .hg sandbox build
```
MihaiTabara: Ah, good fix. I always end up nuking the `build` folder when I run tox locally in a good while.

PR #170, opened 2018-07-11
Support win64-asan-reporter and mac-asan-reporter platforms

choller:
These changes are similar to what catlee did before for Linux to support asan-reporter.

I am currently working on Windows support in bug 1473259 but Mac support is also planned and I figured it is easier to make these changes in one on the beetmover side, so it is less work to deploy.

I am also going to change the m-c patch in that bug to match the platform descriptor used here (win64-asan-reporter).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,14 +6,8 @@ This project adheres to `Semantic Versioning <http://semver.org/>`__.
 
 .. towncrier release notes start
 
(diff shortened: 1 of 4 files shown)


choller:
@MihaiTabara I would very much appreciate that! Please let me know as soon as it is in production, so I can kick off a new try run with the mozilla-central patch.

PR #174, opened 2018-07-25
Add support for language pack checksums files.

Callek:


Callek:
This is in support of https://bugzilla.mozilla.org/show_bug.cgi?id=1463749 and I would love help testing that it produces the checksum files as expected.

MihaiTabara:
We need to fix the tests, I assume the taskcluster 4.0.0 is the culprit at first glance.
In order to test this: we can [roll-out a staging version](https://github.com/mozilla-releng/beetmoverscript/#deploy-a-new-version-to-staging) containing these changes  and trigger some staging releases on maple. I can help with that if needed.

beetmoverscript/newsfragments/1463749.feature
```diff
@@ -0,0 +1 @@
+Add support for language pack checksums.
```
MihaiTabara: @Callek towncrier has a limited amount of operations that it supports. Can you please s,feature,added this file?

beetmoverscript/constants.py
```diff
@@ -151,7 +151,8 @@
 )
 
 CHECKSUMS_CUSTOM_FILE_NAMING = {
-    'beetmover-source': '-source'
+    'beetmover-source': '-source',
+    'release-sign-and-push-langpacks': '-langpack',
```
MihaiTabara: This is the culprit. This should be `release-beetmover-signed-langpacks` instead as the logic from https://github.com/mozilla-releng/beetmoverscript/pull/174/files#diff-21f3f939e3375901322bd3639f4394bcR93 looks into the ['tags']['kind'] which in this case is set to `release-beetmover-signed-langpacks`.

MihaiTabara:
Leftover fixes: 
* we need to fix the requirements to cope with the taskcluster 4.0.0
* rename towncrier file s,feature,added. I can do that do whenver we roll-out, not a biggie.

Callek:
@MihaiTabara this is ready for a new look.

Callek:
This now has a fixed test, @MihaiTabara can you please deploy when you get a chance?

PR #175, opened 2018-07-27
require python3.7 tests to be green

escapewindow:


escapewindow:
Someone's not really afk! :) Thanks!

PR #185, opened 2018-09-27
Declarative artifacts, reworking lguo's PR

srfraser:
This variant removes most of the code duplication. Instead, `push_to_nightly`, `move_beets` and similar can all accept both an old-style `manifest` and a new-style `artifact_map`, with the latter taking priority. These functions know how to extract the right data from both, and lower-level functions like `move_beet` get all the arguments they need without knowing the difference.


beetmoverscript/script.py
```diff
@@ -316,26 +320,46 @@ def list_bucket_objects(context, s3_resource, prefix):
...
+                write_json(source, get_updated_buildhub_artifact(
+                    source, installer_path, context, locale, manifest=manifest, artifact_map=artifact_map))
+
+            if artifact_map:
+                task_id = get_taskId_from_full_path(source)
+                artifact_pretty_name = artifact_map[task_id][artifact]['checksums_path']
+                destinations = artifact_map[task_id][artifact]['destinations']
+                update_balrog_manifest = artifact_map[task_id][artifact].get(
+                    'update_balrog_manifest', False)
+                balrog_format = artifact_map[task_id][artifact].get('balrog_format', '')
```
MihaiTabara: 👍

beetmoverscript/test/templates/fake_nightly.yml
```diff
@@ -43,7 +43,7 @@ mapping:
     target.apk:
       s3_key: fake-{{ version }}.{{ locale }}.target.apk
       destinations:
-        - {{ upload_date }}-{{ branch }}-fake/{{ locale_prefix }}fake {{ version }}.{{ locale }}.target.apk
+        - {{ upload_date }}-{{ branch }}-fake/{{ locale_prefix }}fake-{{ version }}.{{ locale }}.target.apk
```
MihaiTabara: 👍

beetmoverscript/task.py
```diff
@@ -188,21 +188,41 @@ def update_props(context, props, platform_mapping):
...
     # Update fields
     contents['download']['size'] = utils.get_size(installer_path)
     contents['download']['date'] = str(arrow.utcnow())
     contents['download']['url'] = urllib.parse.urljoin(url_prefix, url)
 
     return contents
+
+
+def get_taskId_from_full_path(full_path_artifact):
+    """ Temporary fix: Extract the taskId from a full path artifact
```
MihaiTabara: Good point. We might find a better solution here once we can test things end-to-end.

beetmoverscript/test/test_script.py
```diff
@@ -398,14 +408,14 @@ def test_enrich_balrog_manifest(context, branch, action):
         }
         entry.update(complete_info)
         if partials:
-                entry['partialInfo'] = [
```
MihaiTabara: Sorry for the whitespace/tabs mess, thanks for fixing this!

beetmoverscript/utils.py
```diff
@@ -286,3 +287,11 @@ def get_bucket_name(context, product):
 
 def get_bucket_url_prefix(context):
     return context.config['bucket_config'][context.bucket]['url_prefix']
+
+
+def validated_task_id(task_id):
+    """Validate the format of a taskcluster taskId."""
+    pattern = r'^[A-Za-z0-9_-]{8}[Q-T][A-Za-z0-9_-][CGKOSWaeimquy26-][A-Za-z0-9_-]{10}[AQgw]$'
```
MihaiTabara: I'm wondering if it wouldn't make more sense here to reuse `taskcluster` python package. There must be something there already for validating taskIds. Or maybe even query the queue for the existence of that task? Seems a bit less fragile than a regex that may change in the future with more Taskcluster development. Certainly not blocking now though, just a nit.

srfraser: I don't see that feature in taskcluster-client.py at the moment. Some of their tests use taskIds like `123`, so I'd imagine there's no internal validation in the module, too. 

We could ask them to add it, of course - the regex was taken from https://docs.taskcluster.net/docs/reference/platform/taskcluster-queue/references/api - we could query https://schemas.taskcluster.net/queue/v1/task.json# for the live value if nothing ends up being added to the client.

beetmoverscript/task.py
srfraser: For proper attribution, it was Lisa's comment, I just left it there because I agreed!

beetmoverscript/data/artifactMap_beetmover_task_schema.json
```diff
@@ -0,0 +1,102 @@
...
+                    "uniqueItems": true
+                },
+                "artifactMap": {
+                    "type": "object",
+                  
+                    "minItems": 1,
+                    "uniqueItems": true
+                }
+            },
+            "required": ["upload_date", "upstreamArtifacts", "releaseProperties"],
```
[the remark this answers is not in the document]
srfraser: Yes, you're right.

srfraser:
We also need to improve the coverage before we can merge, unless we bypass that check. I'll see about the state of birch

beetmoverscript/maven_utils.py
```diff
@@ -49,6 +57,11 @@ def _get_maven_expected_files_in_archive(mapping_manifest):
     ]
 
 
+def _get_maven_expected_files_from_map(artifact_map, task_id):
+    mapping = next(entry for entry in artifact_map if entry['taskId'] == task_id)
```
MihaiTabara: Nit: maybe a quick comment to explain the use of `next()` here?

beetmoverscript/script.py
```diff
@@ -83,10 +69,30 @@
...
     context.checksums = dict()
 
-    # for each artifact in manifest
-    #   a. map each upstream artifact to pretty name release bucket format
-    #   b. upload to corresponding S3 location
-    await move_beets(context, context.artifacts_to_beetmove, mapping_manifest)
+    # TODO: if artifactMap passes schema validation
+    if context.task['payload'].get('artifactMap'):
+        # determine artifacts to beetmove
+        context.artifacts_to_beetmove = get_upstream_artifacts(context, preserve_full_paths=True)
```
MihaiTabara: Glorious future when this will be the sole option 🥇

beetmoverscript/script.py
```diff
@@ -316,26 +336,49 @@ def list_bucket_objects(context, s3_resource, prefix):
...
+                destinations = [os.path.join(manifest["s3_bucket_path"],
+                                             dest) for dest in
+                                manifest['mapping'][locale][artifact]['destinations']]
+                update_balrog_manifest = manifest['mapping'][locale][artifact].get(
+                    'update_balrog_manifest')
+                # Adjust old template format.
+                # artifact map specifies these separately.
+                # templates say "update_balrog_manifest": true
+                # or "update_balrog_manifest": {"format": mozinfo}
+                if isinstance(update_balrog_manifest, dict):
```
MihaiTabara: Once declarative artifacts is in place, I'd really like to separate this behavior. I'm not particularly fond of this variable reuse, it makes it complicated to read for someone not familiar with the code. However, this is a carry-over thing and is not blocking existing PR. We'll address it hopefully in H1 2019.

beetmoverscript/task.py
```diff
@@ -188,21 +188,42 @@ def update_props(context, props, platform_mapping):
...
```
MihaiTabara: 👍 for the comment, it's useful to see the end result.

beetmoverscript/test/test_maven_utils.py
```diff
@@ -114,8 +114,120 @@
...
+            get_maven_expected_files_per_archive_per_task_id(
+                upstream_artifacts_per_task_id, mapping_manifest=mapping_manifest)
     else:
         assert get_maven_expected_files_per_archive_per_task_id(
-            upstream_artifacts_per_task_id, mapping_manifest
+            upstream_artifacts_per_task_id, mapping_manifest=mapping_manifest
+        ) == expected
+
+
+@pytest.mark.parametrize('upstream_artifacts_per_task_id, artifact_map, expected, raises', ((
```
MihaiTabara: Awesome tests!

beetmoverscript/test/test_script.py
```diff
@@ -138,6 +141,88 @@ def fake_list(*args):
             await push_to_maven(context)
 
 
+@pytest.mark.asyncio
```
MihaiTabara: Nice!

beetmoverscript/test/test_script.py
```diff
@@ -207,7 +292,8 @@ def fake_filter(**kwargs):
     s3_resource.Bucket = fake_bucket
     bucket.objects.filter = fake_filter
 
-    assert list_bucket_objects(mock.MagicMock(), s3_resource, None) == {"one": "asdf", "two": "foo"}
+    assert list_bucket_objects(mock.MagicMock(), s3_resource, None) == {
```
MihaiTabara: Thanks for fixing lint PEP8 issues too 👍

beetmoverscript/test/test_task.py
```diff
@@ -343,3 +350,24 @@ def test_get_release_props(context, mocker, taskjson, locale, relprops, expected
...
     context.task['tags']['kind'] = kind
     assert is_custom_checksums_task(context) == expected
+
+
+@pytest.mark.parametrize("path, expected", ((
+    '/src/beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/target.mozinfo.json', 'eSzfNqMZT_mSiQQXu8hyqg'
+), (
+    'test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/target.mozinfo.json', 'eSzfNqMZT_mSiQQXu8hyqg'
+),))
+def test_get_taskId_from_full_path(path, expected):
```
MihaiTabara: @srfraser  Can we please add a test for when the string `cot` lies within the task-id itself? Like
`/src/beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSicotu8hyqg/public/build/target.mozinfo.json` should return `eSzfNqMZT_mSicotu8hyqg`

In the function definition, you're splitting based on the os separators, which makes sense. `cot` will be among the indexed items in that separated list - https://github.com/mozilla-releng/beetmoverscript/pull/185/files#diff-21f3f939e3375901322bd3639f4394bcR223 But for the safety, can we please add a test that makes sure the presence of the `cot` substring is not disruptive by any form to this logic?

beetmoverscript/test/test_utils.py
```diff
@@ -14,15 +14,17 @@
...
+                                   extract_file_config_from_artifact_map)
 from beetmoverscript.constants import HASH_BLOCK_SIZE
 
 assert context  # silence pyflakes
 
 
 # get_hash {{{1
 def test_get_hash():
-    correct_sha1s = ('cb8aa4802996ac8de0436160e7bc0c79b600c222', 'da39a3ee5e6b4b0d3255bfef95601890afd80709')
+    correct_sha1s = ('cb8aa4802996ac8de0436160e7bc0c79b600c222',
```
MihaiTabara: Again, thanks for fixing these various linting issues.

beetmoverscript/test/test_work_dir/task.json
```diff
@@ -53,6 +53,25 @@
         }
       ]
     },
+    "extra": {
```
MihaiTabara: AFAIK, this is usefless as it's not processed in beetmover. But most likely you've done a copy-paste from a real task so this certainly doesn't hurt 👍

beetmoverscript/test/test_work_dir/task_artifact_map.json
```diff
@@ -0,0 +1,137 @@
+{
```
MihaiTabara: This is great, thanks for adding this and the follow-up tests! 🥇  Once declarative artifacts is wrapped up, I think the glorious future should hold integration tests. As in, for each type of beetmover tasks, we should have an integration test for (that includes per product too). Unfortunately, this is something skipped from the very beginning of beetmoverscript implementation so the current logic is great and thanks again for adding these!

beetmoverscript/utils.py
```diff
@@ -286,3 +287,22 @@ def get_bucket_name(context, product):
...
+    if re.fullmatch(pattern, task_id):
+        return task_id
+    raise ValueError("No valid taskId found.")
```
MihaiTabara: Should we raise an `TaskVerificationError` instead?

beetmoverscript/utils.py
```diff
@@ -161,7 +161,8 @@ def generate_beetmover_template_args(context):
     elif 'locale' in task['payload']:
         tmpl_args['locales'] = [task['payload']['locale']]
 
-    product_name = get_product_name(release_props["appName"].lower(), release_props["stage_platform"])
+    product_name = get_product_name(
```
MihaiTabara: Last but not least, thanks for fixing these lint issues! 🥇

beetmoverscript/script.py
srfraser: Agreed, it's not easy to read, and it should hopefully disappear with the old manifest templates.

beetmoverscript/test/test_work_dir/task_artifact_map.json
srfraser: +1

beetmoverscript/test/test_task.py
```diff
@@ -354,6 +354,8 @@ def test_is_custom_beetmover_task(context, kind, expected):
 
 @pytest.mark.parametrize("path, expected", ((
     '/src/beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/target.mozinfo.json', 'eSzfNqMZT_mSiQQXu8hyqg'
+), (
+    '/src/beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8cotg/public/build/target.mozinfo.json', 'eSzfNqMZT_mSiQQXu8cotg'
```
MihaiTabara: 👍 thank you 🥇

PR #186, opened 2018-09-27
Beetmove the updater-dep test package into candidates dirs.

mozbhearsum:
This is in support of the dep updater work in https://bugzilla.mozilla.org/show_bug.cgi?id=1490119. I've been working on a patch that generates an "updater-dep" package with a different updater binary in it. If we end up using it, we need to beetmove it into the candidates directory as we do for other test packages.

This is untested, but I expect it will fix failures such as https://treeherder.mozilla.org/#/jobs?repo=maple&selectedJob=201507334.


mozbhearsum:
OK, I managed to do a quick test of this - it seems to work fine. Eg: https://taskcluster-artifacts.net/Hp8GxQptQ8eAtm557dhFBQ/0/public/logs/live_backing.log

mozbhearsum:
We don't need it in l10n, but we do in Devedition, come to think of it.

MihaiTabara:
Okay, sounds good. Let me also resolve the confligts as in the meantime we merged another beetmoverscript PR.

PR #188, opened 2018-10-29
Bug 1442684 - remove bouncer.apk as it seems unused.

MihaiTabara:
Some cleanup whilst working on declarative-artifacts. This file seems unused. Can't seem to find it neither in the release, nor in the candidates section so it could be a relic from first iterations of tcmigration.


MihaiTabara:
Thanks both for review and context.

PR #190, opened 2018-11-02
Support msi mime type

Callek:
This also adds an explicit test if there is no detected Mime Type to make future discovery of this issue easier.

### DIFF
--- a/beetmoverscript/constants.py
+++ b/beetmoverscript/constants.py
@@ -1,5 +1,6 @@
 MIME_MAP = {
     '': 'text/plain',
+    '.apk': 'application/vnd.android.package-archive',
     '.asc': 'text/plain',
     '.beet': 'text/plain',
     '.bundle': 'application/octet-stream',
@@ -8,8 +9,8 @@ MIME_MAP = {
     '.dmg': 'application/x-iso9660-image',
     '.json': 'application/json',
     '.mar': 'application/octet-stream',
+    '.msi': 'application/x-msi',
     '.xpi': 'application/x-xpinstall',
-    '.apk': 'application/vnd.android.package-archive',
 }
 
 STAGE_PLATFORM_MAP = {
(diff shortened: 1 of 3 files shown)


beetmoverscript/constants.py
```diff
@@ -8,8 +9,8 @@
     '.dmg': 'application/x-iso9660-image',
     '.json': 'application/json',
     '.mar': 'application/octet-stream',
+    '.msi': 'application/x-msi',
```
MihaiTabara: @Callek is this the same with `application/x-msdownload`?

beetmoverscript/script.py
```diff
@@ -555,10 +555,16 @@ def enrich_balrog_manifest(context, locale):
...
+                                            path, mime_type
+                                        ))
     api_kwargs = {
         'Bucket': get_bucket_name(context, product),
         'Key': s3_key,
-        'ContentType': mimetypes.guess_type(path)[0]
+        'ContentType': mime_type,
     }
     headers = {
         'Content-Type': mimetypes.guess_type(path)[0],
```
MihaiTabara: Nit: We could reuse `mime_type` here too since we already declared it.

beetmoverscript/test/test_script.py
```diff
@@ -309,6 +309,15 @@ def test_enrich_balrog_manifest(context, branch, action):
     await beetmoverscript.script.upload_to_s3(context, 'foo', 'bar')
 
 
+@pytest.mark.asyncio
```
MihaiTabara: GG for the test!

beetmoverscript/constants.py
Callek: Googling made me only find `/x-msi` as appropriate since /x-msdownload is more for dll and exe http://mimeapplication.net/x-msdownload

I also have /x-msi on my Ubuntu 18.04 `/etc/mime.types`

MihaiTabara: I have `application/x-msdownload` on my macbook but I guess we can keep yours instead. It doesn't matter what we have as long as it's something valid that makes the presigned url function generation happy.  If we have to, we can always address later.

PR #194, opened 2018-11-27
Add snapshots support for maven push.

MihaiTabara:


MihaiTabara:
@JohanLorenzo This is not yet ready for review - I'm more interested in a general-workflow touch base to see if this makes sense. Once this validates and we have artifacts pushed to snapshots repo, I can work on details, add tests and prep this for formal review.

beetmoverscript/utils.py
```diff
@@ -179,14 +179,17 @@ def _generate_beetmover_template_args_maven(task, release_props):
     }
 
     # Change version number to major.minor.buildId because that's what the build task produces
-    payload_version = FirefoxVersion.parse(task['payload']['version'])
+    payload_version = MavenVersion.parse(task['payload']['version'])
```
MihaiTabara: @JohanLorenzo I'm interested particularly of your opinion here.

beetmoverscript/test/test_utils.py
```diff
@@ -326,7 +326,6 @@ def test_beetmover_template_args_generation(context, taskjson, partials):
     context.task['payload']['upstreamArtifacts'][0]['locale'] = 'en-US'
     expected_template_args['template_key'] = 'fake_nightly'
     expected_template_args['locales'] = ['en-US']
-    # import pdb; pdb.set_trace()
```
JohanLorenzo: :+1:

beetmoverscript/utils.py
JohanLorenzo: It makes sense to me. That's a great change. `FirefoxVersion` was there before `MavenVersion` existed. The function is about maven stuff so yeah, let's use it.

beetmoverscript/zip.py
```diff
@@ -180,13 +198,43 @@ def _ensure_all_expected_files_are_present_in_archive(zip_path, files_in_archive
...
+            _ = {
+                'date_timestamp': date,
+                'clock_timestamp': clock,
+                'build_number': bno
+            }
+            # reload the unique_expected_files with their corresponding values
+            # by rendering them via jsone
+            rendered_unique_expected_files = set([jsone.render(f, _) for f in unique_expected_files])
+            identifiers_collection.append(frozenset(_.items()))
+        if file_ not in rendered_unique_expected_files:
```
JohanLorenzo: :warning: This variable is unset if not in snapshot.

beetmoverscript/zip.py
```diff
@@ -180,13 +198,43 @@ def _ensure_all_expected_files_are_present_in_archive(zip_path, files_in_archive
...
+            (date, clock, bno) = _extract_and_check_timestamps(file_,
+                                                               SNAPSHOT_TIMESTAMP_REGEX)
+            _ = {
+                'date_timestamp': date,
+                'clock_timestamp': clock,
+                'build_number': bno
+            }
+            # reload the unique_expected_files with their corresponding values
+            # by rendering them via jsone
+            rendered_unique_expected_files = set([jsone.render(f, _) for f in unique_expected_files])
```
JohanLorenzo: I'm not totally sold on the usefulness of jsone. If I understand correctly, we're using it because:

* we need to render a file at 2 different times.
* Jinja parsing doesn't error out because some params are missing in the context, during the first parsing.

I think jsone is not really the right tool for this job, because:

* jsone is more complex than just for doing variable replacement. It supports [several operators](https://taskcluster.github.io/json-e/#Language/operators) and has its [own syntax](https://taskcluster.github.io/json-e/#Language/expression-syntax).
* jsone syntax might interfere in oddly ways when first parsed by Jinja. E.g.: what would be the output of:

```jinja
{{ {$eval: 'abs(some_number)'} }}
```
?

If we absolutely need to render a jinja template twice, then what do you think of providing a Jinja context like this one, the first time:

```py
context = {
    'date_timestamp': "{{ date_timestamp }}",    # We still don't know what this value is for now. It will be provided at the second rendering step, once we know what the timestamp actually is. 
}
```
?

MihaiTabara:
@JohanLorenzo This is ready for re-review. The only thing leftover to address is the s/`jsone`/`Jinja` change that we've talked about in Mozlando. Wasn't sure if you want that addressed now or later on in a follow-up. I'll branch off this and prep that change once I submit all the `snapshot` patches your way.

MihaiTabara:
@JohanLorenzo Nevermind my above comment, I made the changes in this PR, easier to do it now than later. I went ahead with your proposal and reused Jinja to render twice.

beetmoverscript/zip.py
MihaiTabara: Sounds good, I addressed that in https://github.com/mozilla-releng/beetmoverscript/pull/194/commits/e444f17ed411b80bd0238a89c72100533dff7f14

beetmoverscript/zip.py
```diff
@@ -180,13 +198,45 @@ def _ensure_all_expected_files_are_present_in_archive(zip_path, files_in_archive
...
+    elif len(identifiers_collection_set) == 1:
+        # use this unique identifier per artifacts_id zip archive to munge and
+        # populate the mapping_manifest
+        kwargs = dict(next(iter(identifiers_collection_set)))
+        rendered_mapping = deepcopy(mapping_manifest)
+        for locale, value in mapping_manifest['mapping'].items():
+            rendered_mapping['mapping'][locale] = render_dict(value, kwargs)
+        # munge the original mapping with the rendered counterpart
+        for locale, value in rendered_mapping['mapping'].items():
+            mapping_manifest['mapping'][locale] = rendered_mapping['mapping'][locale]
```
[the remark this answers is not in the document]
MihaiTabara: For some reason I initially thought I'd need to delete some of those entries, hence a mirroring value to change update the `mapping_manifest` object sent by reference. You're absolutely right, great great catch. Thanks!

beetmoverscript/zip.py
```diff
@@ -196,6 +246,45 @@ def _ensure_all_expected_files_are_present_in_archive(zip_path, files_in_archive
     log.info('Archive "{}" contains all expected files: {}'.format(zip_path, unique_expected_files))
 
 
+def render_dict(d, kwargs):
+    def _rendering(s, _):
+        return JINJA_ENV.from_string(s).render(**_)
```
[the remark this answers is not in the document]
MihaiTabara: TIL! 🎖

beetmoverscript/zip.py
```diff
@@ -196,6 +246,45 @@ def _ensure_all_expected_files_are_present_in_archive(zip_path, files_in_archive
     log.info('Archive "{}" contains all expected files: {}'.format(zip_path, unique_expected_files))
 
 
+def render_dict(d, kwargs):
+    def _rendering(s, _):
```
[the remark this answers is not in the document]
MihaiTabara: Let's!

PR #197, opened 2018-12-15
Add win64-aarch64 support.

tomprince:


beetmoverscript/constants.py
```diff
@@ -22,10 +22,7 @@
     'macosx64': 'mac',
     'macosx64-asan-reporter': 'mac-asan-reporter',
     'macosx64-devedition': 'mac',
-    'win32': 'win32',
```
MihaiTabara: These are needed in `task.py` to properly feed the release_props in order to better render the templates, specifically for `platform` value. I might need to look at this with fresh eyes in the morning but how come we're retiring them? Are they not used and I'm missing something?

tomprince:
@MihaiTabara Were you going to do another review pass on this?

beetmoverscript/constants.py
MihaiTabara: Makes sense 👍

PR #200, opened 2018-12-19
Fix buildhub.json download.url, take 2

rail:
Unfortunately the previous fix didn't work properly. Instead of
generating the full URL, it generates only the part after the hostname.
Adding `s3_bucket_path` should fix the issue.

This should also fix the issue, where we try to quote after generating
the full URL when `artifact_map` is specified, so we don't get the
`https%3A//` prefix.


MihaiTabara:
@rail let me know if you want me to roll this out to production.

rail:
> @rail let me know if you want me to roll this out to production.

I'd appreciate this!

PR #202, opened 2018-12-29
Bug 1516542 - Add support for .msi to Thunderbird templates

jfx2006:
Thunderbird version of bug 1475512


beetmoverscript/templates/thunderbird_nightly.yml
```diff
@@ -221,6 +221,12 @@ mapping:
       destinations:
         - {{ upload_date }}-{{ branch }}/thunderbird-{{ version }}.{{ locale }}.{{ platform }}.zip
         - latest-{{ branch }}/thunderbird-{{ version }}.{{ locale }}.{{ platform }}.zip
+    target.installer.msi:
+      s3_key: {{ upload_date }}-{{ branch }}/thunderbird-{{ version }}.{{ locale }}.{{ platform }}.installer.msi
```
MihaiTabara: This should simply be named: `installer.msi`. (The `s3_key` entry is what we'll be shown in the http://archive.mozilla.org/pub/thunderbird/candidates/65.0b3-candidates/build1/SHA256SUMS files as identifiers. No need to have the whole thing.

jfx2006: Done.

PR #204, opened 2019-01-22
Proof of concept metrics collection

srfraser:
For the InfluxDB proof of concept, let's collect some data about upload times.

This will need improvements longer term:

1. The client installation and config should exist in scriptworker, not *script
2. The client currently uses the default destination for udp packets, as that's what we're listening on (localhost:8094) but that may change with our architecture in future, so we should look at putting those in *script config, or perhaps task secrets.
3. I want to get a decorator and context manager for timing so we don't have to work out the duration ourselves everywhere.


requirements-prod.txt
```diff
@@ -5,190 +5,189 @@
...
-    --hash=sha512:97caf11b71a4d80feea02e05a9c18c020aafe3e9b1e2adbd467337fa6095d918b80945831582db50fe6f3aff748875965774579adef4b30b16f82a53fb064a9b \
-    --hash=sha512:abc031b5cd96c59886913582e7226f3c7b9ce4101d11d380f732d16776b990082b18b0c5f91bc098f17f72cbadcf499dfa93069323a8668755c1ce5d06f37f1e \
-    --hash=sha512:d72c52b6ab0a3a77822b48c2db71cd8d54bf55805ce118ce2ebb7b2bce73a6f55611357fbeb15ce66668bcfb1a648dc9176465193971621e6e6e475036138fd6 \
-    --hash=sha512:e370fb65657fe2e13164c3c3f470c665a64312bac9219b35f6b66d0bda3733af433c9acf21e094a5d973d7722f3570b624d955b7778a9e1baac669d7e7aeac61 \
-    --hash=sha512:f29df31f20cf4ae4a185acf8198bc0c582728b81bd258044e754b5fe059a29093743088a165c5bbdf938a2ec359dc4861367c23e160c075eaf912f446c2af41b \
-    --hash=sha512:f59f86c13e08381d3019cc8b5c6a44a49c5f71fa4344ef4c1f1f2fbeee14814e1ddb9a293995ddfd07cf1315bddc9b2de5be7db8673f2f686b912f0e89edde23 \
-    --hash=sha512:fac2e8518ece83e75092a0df25edd13c814a79f2aeec81bfe51bc183b8fd87bcc2986f58c982f5b5f10cf24b2a737bf69d7dce9fe3acf3d996c23f3ff53cdfab \
+    --hash=sha256:1a112a1fdf3802b7f2b182e22e51d71e4a8fa7387d0d38e79a268921b869e384 \
```
MihaiTabara: Thanks for fixing these to `256` 👍

PR #210, opened 2019-04-05
Bug 1539067 - fix buildhub.json logic in declarative artifacts world.

MihaiTabara:


beetmoverscript/utils.py
```diff
@@ -309,6 +309,14 @@ def validated_task_id(task_id):
     raise ValueError("No valid taskId found.")
 
 
+def exists_or_endswith(filename, list_of_artifacts):
+    for artifact in list_of_artifacts:
+        if filename == artifact or filename.endswith(artifact):
```
srfraser: This function gets given `BUILDHUB_ARTIFACT`, which is a string, but it's treating it as a list. So `for artifact in list_of_artifacts` iterates over `['b', 'u', 'i', 'l', 'd', 'h', 'u', 'b', '.', 'j', 's', 'o', 'n']` and will return true for any filename which ends with any of those letters. I think this is why the CI test is failing with json load errors - it's been told to load files that aren't json.

MihaiTabara: Sorry, I forgot to push my local changes, I had fixed that in the meantime.

beetmoverscript/utils.py
```diff
@@ -309,11 +309,38 @@ def validated_task_id(task_id):
     raise ValueError("No valid taskId found.")
 
 
-def extract_file_config_from_artifact_map(artifact_map, path, task_id, locale):
+def exists_or_endswith(filename, basenames):
+    if isinstance(basenames, tuple):
```
srfraser: I'd be tempted to say:
```
    if isinstance(basenames, str):
        basenames = [basenames]
    for artifact in basenames:
```
as that way we catch all iterables (if someone makes a constant a list later on, for example) and we're not duplicating the check.
That said, it works, so not a blocker.

PR #211, opened 2019-04-08
Bug 1542500 Finding the right upstream artifact

srfraser:
In order to translate the full local scriptworker path into an artifact, we
were using `os.path.basename` and then searching through the artifact map,
ignoring locales. This caused the first match to be returned, whether or not it
was correct.

This new variant passes in both the full local path and the upstream artifact
path, so we can match against the full artifact path instead, removing a chance
of duplicate matches.

### DIFF
--- a/beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/buildhub.json
+++ b/beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/buildhub.json
@@ -9,10 +9,10 @@
         "target": "i686-pc-linux-gnu"
     },
     "download": {
-        "date": "2018-11-13T16:34:34.968510+00:00",
+        "date": "2019-04-08T10:19:18.761418+00:00",
         "mimetype": "application/octet-stream",
         "size": 7201,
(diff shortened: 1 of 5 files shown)


beetmoverscript/test/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/buildhub.json
```diff
@@ -9,10 +9,10 @@
         "target": "i686-pc-linux-gnu"
     },
     "download": {
-        "date": "2018-11-13T16:34:34.968510+00:00",
+        "date": "2019-04-08T10:19:18.761418+00:00",
         "mimetype": "application/octet-stream",
         "size": 7201,
-        "url": "https://archive.test/https%3A/archive.test/pub/mobile/nightly/2016/09/2016-09-01-16-26-14-mozilla-central-fake/en-US/fake-99.0a1.en-US.target.apk"
+        "url": "https://archive.test/pub/mobile/nightly/2016/09/2016-09-01-16-26-14-mozilla-central-fake/en-US/fake-99.0a1.en-US.target.apk"
```
JohanLorenzo: 👍 for fixing this test!

MihaiTabara:
Thanks for fixing this!

PR #215, opened 2019-04-16
Bug 1536079 - enable beetmoverscript support for appservices

MihaiTabara:


beetmoverscript/templates/maven_appservices.yml
```diff
@@ -0,0 +1,51 @@
+# This file differs from `maven_geckoview` counterpart as it addresses the
+# appservices which are published under a different group id.
+# Moreover the components lack the javadocs files.
+---
+metadata:
+    name: "Maven repository"
+    description: "Maps artifacts to spec'd maven location"
+    owner: "release@mozilla.com"
+
+s3_bucket_path: maven2/org/mozilla/appservices/{{ artifact_id }}/{{ version }}/
```
JohanLorenzo: I see the bucket path is different, but the rest of the file looks to me. Is it worth combining these 2 files, or declarative artifacts are coming soon enough to not need the refactor?

MihaiTabara: Good point. The latter is the upcoming solution from bug 1541089 in Q2 👍

beetmoverscript/constants.py
```diff
@@ -11,6 +11,10 @@
     '.mar': 'application/octet-stream',
     '.msi': 'application/x-msi',
     '.xpi': 'application/x-xpinstall',
+    '.md5': 'text/plain',
```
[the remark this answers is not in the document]
MihaiTabara: Yes, will address in a follow-up this week. Thanks for raising this, TIL we have the puppet counterpart :P

PR #218, opened 2019-05-17
Support bz2 blobs when using artifact maps.

tomprince:


tomprince:
Looking closer at the code, I think this probably isn't needed.

PR #219, opened 2019-05-21
Bug 1541089 - remove useless templates from beetmoverscript.

MihaiTabara:
Doing some cleanup in beetmoverscript since some of these have already switched to declarative artifacts world.

Leftovers:
* partner repacks
* devedition
* mobile world
* Thunderbird


tomprince:
ESR 60 isn't yet using artifact-maps.

MihaiTabara:
> ESR 60 isn't yet using artifact-maps.

Ah, good point, forgot of esr60. I guess it's safer to let this PR merge only after we'll have branched off esr68.

tomprince:
ESR60 is supported until ESR68.2 is release (so 2019-10-22, per [calendar](https://wiki.mozilla.org/Release_Management/Calendar)), which is several months after ESR68 branches.

MihaiTabara:
@tomprince thanks for following-up here. I will most likely dig into the commits to cherry-pick the DA work to esr60 as well. Moving things to DA is blocking another OKR so I would like to move sooner, rather than wait for esr60 to go EOL. What do you think?

MihaiTabara:
Defunct-ing this in favors of https://github.com/mozilla-releng/beetmoverscript/pull/233

PR #220, opened 2019-05-23
Use `docker push` to publish images

rail:
Until Taskcluster upgrades the ancient version of docker daemon that
supports newer image format (with `manifest.json`), we have to use dind
and `docker push` with all these hacks.

We use something similar for Balrog.


.taskcluster.yml
```diff
@@ -102,12 +102,23 @@ tasks:
             features:
               # Needed for access to secret
               taskclusterProxy: true
+              dind: true
             maxRunTime: 3600
-            image: mozilla/taskboot:latest
+            image: mozillareleases/python-test-runner
```
MihaiTabara: Are we to use this image for all the *scripts that we'll migrate?

rail: I don't have a better candidate for now, but we probably want to use something more generic in the future. I'm going to revisit this approach very soon in any case.

PR #224, opened 2019-06-07
Do not prefix tags

rail:
We already have "v" in the tag


rail:
@escapewindow I bet this request escaped your filters ;) Can you look at it please.

PR #226, opened 2019-06-10
Bug 1548545 - remove zip support in mobile world as we transition to declarative artifacts 

MihaiTabara:


beetmoverscript/test/test_script.py
```diff
@@ -141,88 +141,6 @@ def fake_list(*args):
             await push_to_maven(context)
 
 
-@pytest.mark.asyncio
```
MihaiTabara: I initially attempted to fix these tests up to the point I realized they are no longer needed since we no longer have any zip extraction in the declarative artifacts world. The only assertion that can we could do it so to check that `artifacts_to_beetmove` are correct, but that's slightly redundant since the values stay the same, they are just being reorganized per locale in larger dict. Hence deleting these tests.

MihaiTabara:
The coverage decrease is expected. We can ignore it for now as we'll likely going to remove that whole logic in the next two weeks.

beetmoverscript/script.py
```diff
@@ -176,20 +176,17 @@
 
 async def push_to_maven(context):
     """Push artifacts to locations expected by maven clients (like mvn or gradle)"""
-    artifacts_to_beetmove = task.get_upstream_artifacts_with_zip_extract_param(context)
     context.release_props = get_release_props(context)
     context.checksums = dict()  # Needed by downstream calls
     context.raw_balrog_manifest = dict()    # Needed by downstream calls
 
     if context.task['payload'].get('artifactMap'):
-        context.artifacts_to_beetmove = _extract_and_check_maven_artifacts_to_beetmove(
```
JohanLorenzo: I'm a little bit worried about this change. If we do so, I think we need to land the geckoview changes on all branches, otherwise, it would work. What do you think?

MihaiTabara: Yes, that's correct. I figured since it's mergeday, we have a good window to land and uplift everywhere since there' some expected freeze/breakage timeline. The alternative is to enrich the existing `if` with checking of the `maven.zip` within the `upstreamArtifacts`. We can do that as well.

MihaiTabara: @JohanLorenzo I added another check to allow us a gradual rollout without the need of landing the in-tree patches. Nor the github ones. 

I tested locally against an [old task](https://tools.taskcluster.net/groups/NQDz9u91QgePXkyLj7Tl9A/tasks/e6fUc7noSBigu9i3xYKZVA/details) and a [new task](https://tools.taskcluster.net/groups/DYmBqSrNTWqb7b6_tTL9xw/tasks/WpX4bA1pTgyISs0Qs93O3A/details)

beetmoverscript/test/test_script.py
JohanLorenzo: Nit: I think we can keep these tests while this use-case is still supported.

PR #227, opened 2019-06-11
Support for partner stub installers, and pushing into firefox/releases/partners/

nthomas-mozilla:


nthomas-mozilla:
The idea here is that bouncer script gets a list of partners (in the form <partner>-<sub_config>) which we want to copy from the candidates directory into `pub/firefox/releases/partners/`. We try to match keys in candidates against partners first, because we exclude `partners/` otherwise.  It has to handle the path munging being different for partner builds (source and destination).

Also adds template support for moving stub installers into the candidates dir.

beetmoverscript/script.py
```diff
@@ -153,9 +154,16 @@
         log.warning("Destination {} already exists with {} keys".format(
                     releases_prefix, len(releases_keys_checksums)))
 
-    # Weed out RELEASE_EXCLUDE matches
+    # Weed out RELEASE_EXCLUDE matches, but allow partners specified in the payload
+    push_partners = context.task['payload'].get('partners', [])
     for k in candidates_keys_checksums.keys():
-        if not matches_exclude(k, RELEASE_EXCLUDE):
+        partner_match = get_partner_match(k, push_partners)
+        if partner_match:
```
escapewindow: Should this check some subset of `RELEASE_EXCLUDE` that avoids uploading too many things? Or are we precise about our partner artifacts?

nthomas-mozilla: You're right, this could do with some tightening up.

nthomas-mozilla: I've modified this code so that it does a quick check to see if the key is a possible partner, then checks if the key starts with candidates_dir + 'partner-repack' + any of the publishing partners + 'v1'

nthomas-mozilla:
I'm going to follow up on the regression in test coverage after landing.

PR #232, opened 2019-07-11
Bugfix: geckoview missing actual submission to S3.

MihaiTabara:
Tested locally, works smoothly.


beetmoverscript/script.py
```diff
@@ -192,6 +192,7 @@
                 context.config.get('zip_max_file_size_in_mb', DEFAULT_ZIP_MAX_FILE_SIZE_IN_MB),
                 artifact_map=context.task['payload'].get('artifactMap')
             )
+            await move_beets(context, context.artifacts_to_beetmove, artifact_map=context.task['payload']['artifactMap'])
```
MihaiTabara: Facepalm 🤦‍♂ 😳  Sorry.

JohanLorenzo: I'm sorry too. I should have caught this in review too.

MihaiTabara:
> LGTM. As a follow up, let's add a test that makes sure move beets is called. That will help us in the upcoming code deletion (once we get rid of `target.maven.zip`)

Yeah, good point. I'm adding a todo to myself, but for now I want to roll this to unblock folks there.

PR #234, opened 2019-07-22
generalize worker.json and other parts of dockerization

garbas:


beetmoverscript/__init__.py
```diff
@@ -1,3 +0,0 @@
-from ._version import __version__
-
-__all__ = ['__version__']
```
rail: `__version__` is used by towncrier, but it can be given to it explicitly by specifying `--version`. Can you update README.md accordingly

garbas: done

docker.d/worker.yml
```diff
@@ -0,0 +1,164 @@
...
+  - 'sha256'
+blobs_needing_prettynaming_contents:
+  - 'target.test_packages.json'
+taskcluster_scope_prefix: { "$eval": "TASKCLUSTER_SCOPE_PREFIX" }
+verbose: { "$eval": "VERBOSE == 'true'" }
+zip_max_file_size_in_mb: 300
+bucket_config:
+  $match:
+    'COT_PRODUCT == "firefox" && (ENV == "dev" || ENV == "stage")':
+      dep:
```
garbas: @rail should we rename `dep` to `fake-prod` and `dep-partners` to `fake-prod-partners`?

PR #233, opened 2019-07-18
WIP - initial cleanup in beetmover now that all mobile world has switched to declarative artifacts

MihaiTabara:
Starting off by reverting some of the changes introduced in https://github.com/mozilla-releng/beetmoverscript/pull/194/ and adjust tests going forward.

PR #239, opened 2019-07-31
Bug 1567124 - remove useless templates from beetmoverscript as we switched to DA

MihaiTabara:
@JohanLorenzo Just to double-check, we have declarative artifacts across all trees, right?

I'm doing a more serious cleanup in https://github.com/mozilla-releng/beetmoverscript/pull/233/ but I'm still blocked there by Devedition and some other stuff that's currently riding the trains. But these templates are un-used AFAIK so at least we can remove them for now. Cleanup baby steps :)


MihaiTabara:
Coverage is expected to decrease a bit since we're removing some of the tests. That's okay, we're going to remove the maven associated functions too soon.

MihaiTabara:
Yeah, I confirm, esr60 since uses them. I'm still debating whether it makes sense to uplift patches or just to wait for its EOL in the falls. I'm adding support for Devedition too this week so hopefully there's going to be more templates to kill soon :)

PR #246, opened 2019-08-21
Fix symlink path

rail:


rail:
> I'm guessing you're removing the `apt-get update` on purpose heavy_check_mark

yeah, there is no need to update the indexes, because there is no `apt-get install`

